PR #34, opened 2019-11-22
Refactor release-manager

michaelsauter:
Just moving things around a little.

There are still 2 params required for this quickstarter from the global config. This is not ideal as without those 2 params, we would not need access to the repo at all. And I think with the project-specific cd_user we would like to cut access to `ods-configuration`.

@metmajer Could you share your thoughts on this?


release-manager/custom-create-ocp-resources.sh
```diff
@@ -9,26 +9,29 @@ while [[ "$#" > 0 ]]; do case $1 in
...
+  -r|--repo-base) REPO_BASE="$2"; shift;;
 
   *) echo "Unknown parameter passed: $1"; exit 1;;
 esac; shift; done
 
-jiraUrl=$(cat /data/ods-configuration/ods-project-quickstarters/boilerplates/release-manager/jenkins.env | grep JIRA_URL | awk -F "=" '{print $NF}')
-nexusUrl=$(cat /data/ods-configuration/ods-project-quickstarters/boilerplates/release-manager/jenkins.env | grep NEXUS_URL | awk -F "=" '{print $NF}')
-docgenUrl=$(cat /data/ods-configuration/ods-project-quickstarters/boilerplates/release-manager/jenkins.env | grep DOCGEN_URL | awk -F "=" '{print $NF}')
+jiraUrl=$(cat /data/ods-configuration/ods-core.env | grep JIRA_URL | awk -F "=" '{print $NF}')
+docgenUrl=$(cat /data/ods-configuration/ods-core.env | grep DOCGEN_URL | awk -F "=" '{print $NF}')
```
kiwo: my mistake. shouldnt this be $quickstarterBasepath/ods-config/...  ?

michaelsauter: @kiwo Oh yes, it should be. Are you ok with the other changes? I'd prefer to merge the PR, and then you can update the location?

michaelsauter:
@kiwo This is now ready for review.
@metmajer We talked about this briefly - the quickstarters should not need to access `ods-configuration` anymore. Therefore, I have "hardcoded" `DOCGEN_URL` to be `http://docgen.${PROJECT}-cd.svc:8080`, which will work for the default case. `NEXUS_URL` and `JIRA_URL` I have removed. My understanding is that the MRO lib will handle this gracefully, correct? However, I want to create a follow-up PR in `ods-core` that adds those two env vars to the Jenkins instance on creation.

metmajer:
@michaelsauter this is not correct. `NEXUS_URL` and `JIRA_URL` are required and the release manager will result in a failure in the init phase of the pipeline if those are not in place. As it will be re-inserted in a follow-up PR, I don't see any issues here. Thanks for solving this issue @michaelsauter, I wish I had been able to dedicate more time to finding a solution together!

michaelsauter:
@metmajer Ok, do we need a migration script then for existing Jenkins instances, or are the failures nice enough so that users know what to do?

michaelsauter:
Follow-up PR is at https://github.com/opendevstack/ods-core/pull/304.

metmajer:
This is what users would see. Here, `JiraService` cannot anticipate that the error is because of a missing or empty environment variable.

```
JiraService(String baseURL, String username, String password) {
    if (!baseURL?.trim()) {
        throw new IllegalArgumentException("Error: unable to connect to Jira. 'baseURL' is undefined.")
    }
    ...
```

Issue #12, opened 2019-10-25
Quickstarter-specific memory quotas


michaelsauter:
**Is your feature request related to a problem? Please describe.**
At the moment there is only one memory request/limit for every quickstarter. That is not ideal because different types have different memory requirements. As we have just one setting, that setting is the highest requirement (I guess for be-spring-boot?).

**Describe the solution you'd like**
Have the quickstarter define its memory requirements.

FYI @kiwo maybe you can keep that in mind when you do the ODS 2 changes - it would be nice if this fits into the new structure

michaelsauter:
I think the most straightfoward solution right now would be to just hardcode in the provisioning `Jenkinsfile` what should be rendered inside the component template.

michaelsauter:
The new memory quotas should keep a 1:2 ratio between request/limit. I think that is a typical ratio ...

PR #35, opened 2019-11-22
Implement quickstarter-specific memory quotas

henrjk:
Early version for #12 - Not meant to pull in yet

The changed create-component.sh script checks whether there is an file called 

1. `qs-$NAMESPACE.env`
2. `qs.env'

For the first one it finds it adds a `--param-file` parameter to the tailor invocation. 

@michaelsauter @kiwo let me know in case you'd like to see something different here including the naming of the env files.

If you express that it makes sense I would then go on to find sensible defaults for all quickstarters.


be-golang-plain/ocp.env
```diff
@@ -0,0 +1,2 @@
+MEMORY_LIMIT=64Mi
+MEMORY_REQUEST=32Mi
```
michaelsauter: Please add newlines.

common/scripts/create-component.sh
```diff
@@ -114,6 +118,15 @@ for devenv in dev test ; do
         "--param=ENV=${devenv}"
         )
 
+    if [ ! -z "${QSBASE}" ]; then
```
michaelsauter: Wondering if it should fail or at least print a warning? I guess this is kind of a required param now?

common/scripts/create-component.sh
```diff
@@ -114,6 +118,15 @@ for devenv in dev test ; do
...
+      for envfile in "${QSBASE}/ocp-$devenv.env" "${QSBASE}/ocp.env" ; do
```
michaelsauter: What could the usecase for different config for dev/test be? Not sure I would add this distinction without a usecase ...

be-golang-plain/ocp.env
michaelsauter: Also please note that it seems that each component must have at least 100Mi memory request.

henrjk: Done this.

common/scripts/create-component.sh
henrjk: I will make another adjustment for that.

henrjk:
Added new rebased commits and also different quotas for some of the quickstarters. 

The values meant to:
1. Be on the low end to overburden a cluster when users explore quickstarters.
2. Should work out of box and not be already extremely close to the limit. 
3. Be above 100Mi (128Mi specifically) which our production has as a minimum.

The specific numbers are gut numbers which not have been tested except for the golang quickstarter.

common/scripts/create-component.sh
henrjk: This is now addressed as well.

fe-vue/ocp.env
```diff
@@ -0,0 +1,2 @@
+MEMORY_LIMIT=256Mi
```
michaelsauter: I think all the `fe-*` ones can actually have a `MEMORY_LIMIT=128Mi` as they use Nginx and will likely not need that much memory ...

henrjk: You are suggesting limit and request being 128 for fe-* right?
@akhilsoman can you take a look?

common/scripts/create-component.sh
henrjk: When not provided the quickstarter falls back to the default defined in the yml template. 

I would add a warning unless you feel it should be required. I am 50/50 on this. Let me know if you have a preference.

henrjk: I pushed a comment so this is now a required parameter. One point less variability reduces complexity. Thanks for the feedback
But internal quickstarters which use the scripts will now need to adjust, which should be fine as this is all new.

fe-vue/ocp.env
henrjk: I added the parameter to the other Jenkinsfile(s) now as well.
Regarding changes for the memory limit I wanted to wait for @akhilsoman's input.
On the other hand if the fe quickstarters never use that much memory, there should be no difference to the cluster as they presumably would only use the requested value.

akhilsoman: As @michaelsauter said, all the fe-* quickstarters uses Nginx to serve static content and  it's unlikely that they will need more than 128. But in my opinion, we could keep the limit at 256Mi as the additional memory will not be used from the cluster, unless needed by the container.

michaelsauter: @henrjk Your call.

Let me know when this PR is final for review/merge.

henrjk:
This should now be ready for review and implements the following:

create-component.sh requires a new parameter (`-qs`|`--qsbasepath`) which needs to point to the base directory of the quick starter (either relative to the current work directory or absolute).

create-component.sh will check whether there is a file called:
 `ocp.env` at the -qs directory and if so adds it as a `--param-file` parameter to the tailor invocation.

The Jenkinsfiles using the create-component.sh script in this repository have been adjusted to pass in the `-qs` parameter.

Specific memory limits/request are defined as follows:

```bash
$ find . -name "ocp.env" -print -exec cat {} \;
./fe-angular/ocp.env
MEMORY_LIMIT=256Mi
MEMORY_REQUEST=128Mi
./docker-plain/ocp.env
MEMORY_LIMIT=128Mi
MEMORY_REQUEST=128Mi
./fe-react/ocp.env
MEMORY_LIMIT=256Mi
MEMORY_REQUEST=128Mi
./fe-ionic/ocp.env
MEMORY_LIMIT=256Mi
MEMORY_REQUEST=128Mi
./fe-vue/ocp.env
MEMORY_LIMIT=256Mi
MEMORY_REQUEST=128Mi
./be-typescript-express/ocp.env
MEMORY_LIMIT=256Mi
MEMORY_REQUEST=128Mi
./be-golang-plain/ocp.env
MEMORY_LIMIT=128Mi
MEMORY_REQUEST=128Mi
```
I would propose to check with the maintainers of be-java-springboot, be-python-flask, be-scala-akka, ds-jupyter-notebook, ds-rshiny,  e2e-cypress in separate tickets.

henrjk:
close #12

henrjk:
Resolved merge conflicts.

michaelsauter:
I would prefer to cover all QS in this PR.

@gerardcl Can you give hints for be-python-flask, ds-jupyter-notebook, ds-rshiny

@oalyman Can you give hints for be-scala-akka?

@herrkoch / @akhilsoman Can you give hints for e2e-cypress?

For be-java-springboot, please do 512Mi/1Gi.

herrkoch:
For e2e-cypress I guess take the same memory limits as for fe-angular, but maybe @cschweikert knows more.

michaelsauter:
@herrkoch @akhilsoman I think you were using some headless chrome lately, no? And had issues with memory?

cschweikert:
@herrkoch e2e-cypress doesn't run anything on OpenShift directly. Test runs are happening during a Jenkins build. I guess therfore the environment file is not needed at all.

henrjk:
In one of our projects I am seeing between 200 and 300 Mi for Python. @gerardcl what do you think? 
Otherwise I would propose to set it to 256/512Mi for be-python-flask. Although this is the default I would make it explicit.

gerardcl:
@henrjk yes, let's make it 256/512Mi for `be-python-flask`.
I will provide it soon

gerardcl:
Hi!
I have two questions:
- why not providing CPU limits too?
- shouldn't we also provide limits in the buildconfigs?

michaelsauter:
@gerardcl Valid points, but I'd say unrelated here. This PR focuses only on runtime memory quota.

michaelsauter:
@henrjk Use 1Gi/2Gi for Scala.

henrjk:
This should be complete now.

Issue #8, opened 2019-11-20
be-typescript-express: node version in deployment image doesn't match build image


tbugfinder:
**Describe the bug**
The project is built using nodejs8 however the deployment is based on nodejs6 - this is a mismatch and doesn't work.

https://github.com/opendevstack/ods-project-quickstarters/blob/master/boilerplates/be-node-express/Jenkinsfile#L25

**To Reproduce**
1. Build a project which makes use of modules.

**Expected behavior**
Built modules should be reusable within deployment.

**Screenshots**
<none>

**Affected version (please complete the following information):**
 - OpenShift: [3.11]
 - OpenDevStack [1.2.x]

**Log Output (ensure to remove any confidential information like tokens, project names, etc.**
```
Error: Missing binding /node/node_modules/node-sass/vendor/linux-x64-48/binding.node
Node Sass could not find a binding for your current environment: Linux 64-bit with Node.js 6.x

Found bindings for the following environments:
  - Linux 64-bit with Unsupported runtime (57)

This usually happens because your environment has changed since running `npm install`.
Run `npm rebuild node-sass` to build the binding for your current environment.
    at module.exports (/node/node_modules/node-sass/lib/binding.js:15:13)
    at Object.<anonymous> (/node/node_modules/node-sass/lib/index.js:14:35)
    at Module._compile (module.js:577:32)

```

**Additional context**
Very easy to fix :-)

PR #36, opened 2019-11-22
Use Node 10 for bulid and deployment environment for be-typescript-express qs

akhilsoman:
Deployment image doesn't match build image & Linting error fix for be-typescript-express quick starter. 
Fixing #8


akhilsoman:
@michaelsauter

PR #47, opened 2019-11-26
Fix/shiny and jupyter

### LINKED ISSUE #46: ds-rshiny and ds-jupyter-notebeook: ready for ODS2
Both `ds-rshiny` and `ds-jupyter-notebook` requires getting latest changes and some refactor on the provisioning side.

To get the changes from https://github.com/opendevstack/ods-project-quickstarters/issues/424.
And, to get independent from ods-configuration (enhance the provisioning)

**Additional context**
If we do not apply this changes the provisioning will be failing!

gerardcl:
closes  #46


common/scripts/create-rshiny.sh
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/env bash
```
michaelsauter: @gerardcl Could you move that script out of `common`? I would put this inside the rshiny folder.

gerardcl: @michaelsauter shall I move out the Tailor-related files from common too?

michaelsauter: `common` should only contain things that at least two quickstarters use I'd say. So if there are files that are ONLY used by rshiny, the files should be in rshiny.

gerardcl: then we keep that in common, since that is used for jupyter too, but will rename it

michaelsauter: Haha. Cool. My recent PR #49 introduces some merge conflicts now - should be easy to resolve though ...

Issue #30, opened 2019-11-22
fe-ionic: Uses jenkins-slave-maven for provisioning


michaelsauter:
It should use the same slave that it uses for building (nodejs10).

michaelsauter:
@ManuelFeller If I got your right, also this quickstarter works with nodejs10, could you please open a PR for it? That would be great!

PR #49, opened 2019-11-26
Pass image and ref

### LINKED ISSUE #41: Pass image tag and Git ref as params
The `Jenkinsfile.template` references a specific branch of the sharedd library (e.g. `production`) and an image tag (e.g. `latest`) of a Jenkins slave.

Those have to be customised with every release which is annoying.

I've come up with the idea to make those references variables, as the file is rendered anyway. I'd pass down the value from the provisioning app (which can take it from configuration).

@kiwo @gerardcl Unless you have strong feelings against this, I'd implement that this week.

michaelsauter:
Closes #41.

### DIFF
--- a/be-airflow/Jenkinsfile
+++ b/be-airflow/Jenkinsfile
@@ -1,4 +1,6 @@
 // build params
+def odsImageTag = env.ODS_IMAGE_TAG ?: 'latest'
+def odsGitRef = env.ODS_GIT_REF ?: 'production'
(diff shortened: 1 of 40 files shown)


michaelsauter:
Fixes #30. FYI @ManuelFeller

michaelsauter:
FYI I need to do further adjustments to the release-manager. I'm going to do that in a separate PR.

PR #63, opened 2019-12-03
Feature/be airflow ods2 ready

### LINKED ISSUE #60: be-airflow ready for ODS2
We need to make current airflow in ods-project-quickstarters ready for ODS2.

This means:
- Assure new boilerplate structure works (provisioning)
- Assure we integrate base images in the quickstarter (provisioning also base images in project-cd)

No alternative

**Additional context**
This will be a final move for getting rid of `ods-core/shared-images` and the related `shared-services` namespace

### LINKED ISSUE #76: Rename Airflow QuickStarter
As a suggestion, I would name the airflow QuickStarter as just `airflow` or `airflow-cluster` (what this actually is) and not be-airflow because it gives a false idea that it is a backend component.

gerardcl:
closes  #60 
closes #76 
relates to https://github.com/opendevstack/ods-core/issues/289


gerardcl:
fully tested with success! :) ready to go! 
@hugowschneider

gerardcl:
I will update it soon again since I want to assure it is documented that users will need to update the configmap with right API and CONSOLE URLs

tbugfinder:
As the project used to use prefixes like be-, fe-, ds-, then this component should/could also find a prefix.
How about wf-airflow or wf-airflow-cluster ?

gerardcl:
hi @tbugfinder ! yes! I also thought about that, but one can also say that there are quickstarters hard to categorize, since they can be use-case-wider or unique (e.g.: see also `docker-plain` or `release-manager` ones).

PR #69, opened 2019-12-03
Remove dependency on ods-quickstarters.env file

michaelsauter:
Content of that file right now is https://github.com/opendevstack/ods-core/blob/master/configuration-sample/ods-quickstarters.env.sample. The r-shiny stuff is not needed anymore, right @gerardcl? And the release-manager config will be done differently, see #34.


michaelsauter:
@gerardcl Updated to resolve the merge conflict. Can you please approve? Thanks!

michaelsauter:
@gerardcl Yes but the long-term solution for this is probably to pass them from the prov-app.

PR #78, opened 2019-12-06
ds-dash quickstarter

### LINKED ISSUE #77: DS-Dash Quickstarter
As a data scientist, I would like to rapidly prototype solutions and engage stakeholders as quickly as possible, utilizing a dash board created using plotly dash. This will allow the FE Engineers to focus on CSS ;)

gerardcl:
closes #77

### DIFF
--- a/docs/modules/ROOT/pages/ds-dash.adoc
+++ b/docs/modules/ROOT/pages/ds-dash.adoc
@@ -1,0 +1,92 @@
+= Data Science Plotly Dash (ds-dash)
+
+The project supports generation of Plotly Dash project boilerplate and quick
+installation and integration of it with OpenShift CD pipelines.
+
+== Purpose of this quickstarter
+
+The quickstarter is provisioning a Dash boilerplate ready to be used.
+The package allows easily build a Dash project, using different Python dependencies.
+It contains the basic setup for Docker, Jenkins, SonarQube and OpenShift.
+
+== What files / architecture is generated?
+
+ ├── Jenkinsfile - This file contains Jenkins build configuration settings
+ ├── files
+ │   ├── docker - This folder contains Docker configuration settings
+ │   │   ├── Dockerfile
+ │   │   └── run.sh - This bash script solves issue with permissions for a container user
+ │   └── src
+ │       ├── app.py - This file is the main entry point in the project.
+ │       └── requirements.txt - This file contains a list of required Python modules to run application
+ └── sonar-project.properties - This file contains SonarQube configuration settings
+
+== Frameworks used
+
+* https://docs.python.org/3.7/tutorial/[Python 3.7]
+* https://plot.ly/dash/[Dash]
+
+== Usage - how do you start after you provisioned this quickstarter
+
+The project should be started automatically by OpenShift. Server should be started
+on the port 8080 in the debug mode.
+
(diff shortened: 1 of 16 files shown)


docs/modules/ROOT/pages/ds-dash.adoc
```diff
@@ -0,0 +1,92 @@
...
+
+The project should be started automatically by OpenShift. Server should be started
+on the port 8080 in the debug mode.
+
+[source,python]
+----
+app.run('0.0.0.0', 8080, debug=True)
+----
+
+To disable a debug mode set debug to *False*.
```
sklingel: I would rather suggest to write instructions how to start this thing using docker. Don't think it makes much sense to install everything locally, when it needs to run in docker later , anyway

ds-dash/files/docker/Dockerfile
```diff
@@ -0,0 +1,73 @@
+FROM python:3.7.0-slim
```
sklingel: Why do we not go over centos or redhat? I thought this the way to go?

ds-dash/files/docker/run.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+if ! whoami &> /dev/null; then
+  if [ -w /etc/passwd ]; then
+    echo "${USER_NAME:-default}:x:$(id -u):0:${USER_NAME:-default} user:${HOME}:/sbin/nologin" >> /etc/passwd
+  fi
+fi
+
+cd src && exec python3 app.py
```
sklingel: Why not start it with gunicorn directly? Since dash is based on flask, we face the same issues as with the other quickstarters

gerardcl: I do not see this quickstarter something that is going to be a production ready app, but more a fast prototyping tool for fast interactions and testings with users.

ds-dash/files/docker/Dockerfile
gerardcl: We had a conversation in the past in another thread and users prefer using official images rather than self-backed, which are indeed harder to maintain. Anyway, one can always switch the defaults to any custom requirement.

ds-dash/files/docker/run.sh
sklingel: Does it have any downside using gunicorn? Because I definitely see and already have use-cases where dash could/will be used "productive"

docs/modules/ROOT/pages/ds-dash.adoc
gerardcl: This was discussed while creating the quickstarter. The agreement was to locally just run python, so no need for any docker when developing locally. Then, it does not mean that if one ends up requiring, for example, especial dependencies that require OS libraries one will need to go over docker. But this is definitely a wide use case (e.g.: requiring especial OS distros, special dependencies and compilation libraries, other python package management tools...). That is why we wanted to keep it simple.

ds-dash/files/docker/run.sh
gerardcl: let's meet and define what does here mean "productive"

michaelsauter:
Closing this for now. Please continue the effort internally, and we can revisit adding this officially for v3.

PR #80, opened 2019-12-09
NEXUS_HOST cleanup

### LINKED ISSUE #70: Airflow requires NEXUS_HOST in custom-create-component.sh but does not use it
We should be able to just remove that param?

gerardcl:
fixes #70


be-java-springboot/Jenkinsfile
```diff
@@ -1,101 +1,92 @@
...
-          }
-      }
-}
-
-
-
-
-
-
+// build params
```
gerardcl: this file is fully changed since it was converted back to LF

michaelsauter:
Originally, Nexus was needed in the BuildConfig, as that was the place where everything was assembled. No longer needed.

PR #83, opened 2019-12-11
Specify CPU constraints

michaelsauter:
* There does not seem to be a 2:1 ratio for CPU limits/requests
* Java/Scala need lots of CPU during startup, that's why their limit is high
* `50m` is set as the minimum as pods won't start with lower values

FYI @gerardcl Do you think you could contribute / add similar things for the data science related quickstarters that do not use the common template?


michaelsauter:
@gerardcl Awesome!

PR #84, opened 2019-12-12
Fix release manager provisioning

michaelsauter:


release-manager/ocp-config/cd-docgen.yml
```diff
@@ -3,6 +3,25 @@ kind: Template
...
+  - name: BITBUCKET_URL
+    displayName: Application
+    description: The URL of your bitbucket server.
+    required: true
+  - name: MEMORY_LIMIT
+    displayName: Memory Limit
+    description: Maximum amount of memory available for the container.
+    value: 512Mi
+    required: true
+  - name: MEMORY_REQUEST
```
gerardcl: we could consider adding CPU limits also

metmajer: Don't be overly aggressive on both memory and CPU unless you understand the usage patterns, please. Have you tried out these values with some actual documents being generated?

metmajer:
@gerardcl @michaelsauter not so sure, see above

release-manager/ocp-config/cd-docgen.yml
gerardcl: shall then be increased to 1Gi? do you know if it has gone over 1Gi usage when using it?

metmajer: We haven't measured this in detail yet, as the Groovy based implementation will be replaced with a Golang one in v3 anyway. Anyway, we should avoid anything that impacts service availability until we have that (see below). I'll look up the current settings on the OCTEST cluster.

gerardcl:
so I guess we could leave this quickstarter without limits but keep an eye on it

michaelsauter:
@metmajer I did not change the memory settings at all here, just fixing the pipeline. 1.2.x had the same constraints: https://github.com/opendevstack/ods-project-quickstarters/blob/master/ocp-templates/ocp-config/cd-docgen/cd-docgen.yml#L108-L117.

Anyway - just let me know what values you want to see for memory and I'll make the change.

metmajer:
Thanks, @michaelsauter and @gerardcl. I misunderstood the impact of those settings then. Don't see a risk then for the moment.

Issue #96, opened 2019-12-28
Jupyter Notebook quickstarter defines no Jupyter version


metmajer:
**Describe the bug**
A quickstarter should bind a technology to a specific version and be transparent about it. The Jupyter quickstarter's only refererence to Jupyter is in `docker/requirements.txt`, where no version is mentioned, see [here](https://github.com/opendevstack/ods-quickstarters/blob/master/ds-jupyter-notebook/files/docker/requirements.txt).

**Expected behavior**
A version shall be added to `src/requirements.txt`.

metmajer:
@kiwo a reminder to pin the version of Flask in your quickstarter. Thanks!

Issue #95, opened 2019-12-28
TypeScript quickstarter defines no TypeScript version


metmajer:
**Describe the bug**
A quickstarter should bind a technology to a specific version and be transparent about it. The quickstarter installs TypeScript via a Yeoman generator in the `Jenkinsfile`, see [here](https://github.com/opendevstack/ods-quickstarters/blob/master/be-typescript-express/Jenkinsfile#L44). This generator, via https://www.npmjs.com/package/generator-node-express-typescript, however, does not mention the Typescript version used.

**Expected behavior**
A version shall be made explicit.

metmajer:
@rattermeyer a reminder to pin the version of TypeScript in your quickstarter. Thanks!

michaelsauter:
@sino92 Assigning to you since you expressed you want to look after it for the 3.0 cycle.

michaelsauter:
Please make sure to update to the latest version if possible.

sino92:
Revisiting a more fundamental question about the quickstarter first: https://github.com/opendevstack/ods-quickstarters/issues/315

sino92:
With the latest remaking of Typescript quickstarter, we have clarity on the used lib versions here in [Jenkinsfile](https://github.com/opendevstack/ods-quickstarters/blob/master/be-typescript-express/Jenkinsfile#L17) and in [ODS documentation](https://github.com/opendevstack/ods-quickstarters/blob/master/docs/modules/quickstarters/pages/be-typescript-express.adoc).
Closing this issue now.

Issue #94, opened 2019-12-28
Python Flask quickstarter has no Flask version


metmajer:
**Describe the bug**
A quickstarter should bind a technology to a specific version and be transparent about it. The Flask quickstarter's only refererence to Flask is in `src/requirements.txt`, where no version is mentioned, see [here](https://github.com/opendevstack/ods-quickstarters/blob/master/be-python-flask/files/src/requirements.txt).

**Expected behavior**
A version shall be added to `src/requirements.txt`. Suggestion: latest stable is `1.1.1`, according to the project's [home](https://www.palletsprojects.com/p/flask/).

metmajer:
@kiwo You can refer to the following [requirements.txt](https://github.com/opendevstack/ods-quickstarters/blob/master/ds-ml-service/files/src/requirements.txt) on how to set a version (mind though that this one is not using the latest stable of Flask).

metmajer:
@kiwo a reminder to pin the version of Flask in your quickstarter. Thanks!

PR #97, opened 2019-12-29
WIP: Add quickstarter metadata

### LINKED ISSUE #75: Provide quickstarter metadata in release-manager.yml
To provide ods component properties in code
we would need the following addtional informations in the release-manager.yml file:

- Identification
- Name of Software, 
- Supplier, 
- Version, 
- Description of Functionality, 
- References

Those shall be used in the generation of "System Design Specification" Document (DSD) section 5.2

metmajer:
Closes #75.

Issues have been created for quickstarters which do not make use of explicit versioning for their specific technology. For the time being, versions have been documented as `latest`.

- Flask: https://github.com/opendevstack/ods-quickstarters/issues/94
- Jupyter Notebook: https://github.com/opendevstack/ods-quickstarters/issues/96
- TypeScript: https://github.com/opendevstack/ods-quickstarters/issues/95

@hugowschneider @sklingel @gerardcl: I applied version `1` to the `ds-ml-service` for not knowing better. This attribute is required, so please feel free to provide a more suitable value.

@michaelsauter The version fields in the `release-manager.yml` files will need to be maintained when a quickstarter's dominant technology version is bumped. Can we add this as a requirement to the contribution guidelines of this project?

**TODO**
- [x] Add `release-manager.yml` description to each quickstarter's .adoc file
- [x] Add a note to `plain-docker` quickstarter's .adoc file saying that entries in `release-manager.yml` need to be provided manually and are required if delivery by the Release Manager

### DIFF
--- a/be-python-flask/files/release-manager.yml
+++ b/be-python-flask/files/release-manager.yml
@@ -1,2 +1,8 @@
 ---
+metadata:
+  name: Flask
+  description: Flask is a micro web framework written in Python.
+  supplier: https://www.palletsprojects.com/p/flask/
+  version: latest
+
 dependencies: []
--- a/be-python-flask/files/release-manager.yml
+++ b/be-python-flask/files/release-manager.yml
@@ -1,8 +1,2 @@
 ---
-metadata:
-  name: Flask
-  description: Flask is a micro web framework written in Python.
-  supplier: https://www.palletsprojects.com/p/flask/
-  version: latest
-
 dependencies: []
--- a/airflow-cluster/base-images/README.md
+++ b/airflow-cluster/base-images/README.md
@@ -1,6 +1,6 @@
 # Airflow base images
 
-Images used in [ods-quickstarters/airflow-cluster](https://github.com/opendevstack/ods-quickstarters/airflow-cluster)
+Images used in [ods-quickstarters/airflow-cluster](https://github.com/opendevstack/ods-quickstarters/tree/master/airflow-cluster)
 
 ## Contents
 
--- a/airflow-cluster/files/metadata.yml
+++ b/airflow-cluster/files/metadata.yml
@@ -1,0 +1,6 @@
+---
+metadata:
+  name: Apache Airflow
+  description: "Airflow is a platform created by community to programmatically author, schedule and monitor workflows. Technologies: Airflow 1.10, Python 3.6"
+  supplier: https://airflow.apache.org
+  version: 2.x
--- a/airflow-cluster/files/release-manager.yml
+++ b/airflow-cluster/files/release-manager.yml
@@ -1,8 +1,2 @@
 ---
-metadata:
-  name: Apache Airflow
-  description: Airflow is a platform created by community to programmatically author, schedule and monitor workflows.
-  supplier: https://airflow.apache.org
-  version: 1.10.3
-
 dependencies: []
--- a/airflow-cluster/files/release-manager.yml
+++ b/airflow-cluster/files/release-manager.yml
@@ -1,2 +1,8 @@
 ---
+metadata:
+  name: Apache Airflow
+  description: Airflow is a platform created by community to programmatically author, schedule and monitor workflows.
+  supplier: https://airflow.apache.org
+  version: 1.10.3
+
 dependencies: []
--- a/be-golang-plain/files/metadata.yml
+++ b/be-golang-plain/files/metadata.yml
@@ -1,0 +1,6 @@
+---
+metadata:
+  name: Go
+  description: "Go is an open source programming language that makes it easy to build simple, reliable, and efficient software. Technologies: Go 1.12"
+  supplier: https://golang.org
+  version: 2.x
--- a/be-golang-plain/files/release-manager.yml
+++ b/be-golang-plain/files/release-manager.yml
@@ -1,8 +1,2 @@
 ---
-metadata:
-  name: Go
-  description: Go is an open source programming language that makes it easy to build simple, reliable, and efficient software.
-  supplier: https://golang.org
-  version: 1.12
-
 dependencies: []
--- a/be-golang-plain/files/release-manager.yml
+++ b/be-golang-plain/files/release-manager.yml
@@ -1,2 +1,8 @@
 ---
+metadata:
+  name: Go
+  description: Go is an open source programming language that makes it easy to build simple, reliable, and efficient software.
+  supplier: https://golang.org
+  version: 1.12
+
 dependencies: []
--- a/be-java-springboot/files/metadata.yml
+++ b/be-java-springboot/files/metadata.yml
@@ -1,0 +1,6 @@
+---
+metadata:
+  name: OpenJDK
+  description: "OpenJDK is a free and open-source implementation of the Java Platform, Standard Edition. Technologies: Spring Boot 2.1, OpenJDK 11"
+  supplier: https://adoptopenjdk.net
+  version: 2.x
--- a/be-java-springboot/files/release-manager.yml
+++ b/be-java-springboot/files/release-manager.yml
@@ -1,8 +1,2 @@
 ---
-metadata:
-  name: OpenJDK
-  description: OpenJDK is a free and open-source implementation of the Java Platform, Standard Edition.
-  supplier: https://openjdk.java.net
-  version: 11
-
 dependencies: []
--- a/be-java-springboot/files/release-manager.yml
+++ b/be-java-springboot/files/release-manager.yml
@@ -1,2 +1,8 @@
 ---
+metadata:
+  name: OpenJDK
+  description: OpenJDK is a free and open-source implementation of the Java Platform, Standard Edition.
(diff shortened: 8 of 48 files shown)


metmajer:
@clemensutschig @michaelsauter Great! I want to provide the remaining items under TODO before I merge, so we can consider this PR to be feature complete. For the other described activities, I'll create issues as you suggested.

be-python-flask/files/release-manager.yml
```diff
@@ -1,2 +1,8 @@
 ---
+metadata:
+  name: Flask
+  description: Flask is a micro web framework written in Python.
+  supplier: https://www.palletsprojects.com/p/flask/
+  version: latest
```
michaelsauter: Just realized that in `be-java-springboot` the version is `11`, referring to the JVM. Here, you've specified `latest`, which I assume refers to the Flask version, as the Python version is actually fixed in the quickstarter.

@metmajer Now I am wondering if this version is supposed to be for the language or for the framework if the quickstarter has both language and framework in it?

metmajer: I have tried to identify a dominant technology. Users here "buy" Flask, not Java. Hence the reasoning. Of course, this reasoning has to be shared. Thoughts?

clemensutschig: We should add the Framework version

segator: Agree with you Martin,
$primaryToolFramework-($secondary/environment)?
Example for spring
Multiple tags need to be done to define defaults cases
* latest  <- **default to latest stable?**
* 5.1.6  <-  **default to openjdk11?**
* 5-oraclejdk12
* 5.1.6-openjdk11
* 5.1-openjdk11

The problem is when you have multiple tech in the same image.
JDK + Groovy  + Spring?
we can follow same syntax but tags name will be so long.
* 5.1.6-groovy3.0-openjdk11 ?
Maybe ignore long tags and set all tech versioning as labels in the image.

metmajer: `latest` needs to be avoided, as otherwise, the version (and expected features) is undefined. As indicated above, members have been advised to provide concrete versions.

To me, there is no sense in selling the JVM version with the product, and a version for the framework is enough. Simplicity :)

metmajer:
@michaelsauter @clemensutschig @segator the open points have been added. Ready when you are.

metmajer:
Just for reference, the quickstarter descriptions have been taken from the respective official sites.

be-python-flask/files/release-manager.yml
metmajer: @michaelsauter I want to make sure we are in agreement before I merge

michaelsauter: @metmajer I agree with your reasoning, but shouldn't the version for `be-java-springboot` be `2.2` then (for Spring Boot as the framework) instead of `11`?

metmajer: Ah, now I've got you. I was too focused on the Flask snippet above. In fact, that's a tricky one, since I personally would eye more on the Java 11 than on Spring Boot 2.2. Would a combined version be appropriate here?

georgfedermann: as a newcomer with a fresh view on things, I would intuitively expect the version to be the quickstarter's own version, and expect to find information about frameworks, languages, build tools, etc. in the description.
E.g.
name: be-spring-boot
description: spring-boot **2.1.10**, OpenJDK 11, maven or gradle something
supplier: spring.io
version: **2**
if the versions of one of the technologies gets bumped, so does the quickstarter's version, as in
name: be-spring-boot
description: spring-boot **2.2**, OpenJDK 11, maven or gradle something
supplier: spring.io
version: **3**
That way, the quickstarter's version could also reflect any other change within the quickstarter other than upgrades of its major technologies. 
Otherwise, if we quote the version of the main technology in the quickstarter version and upgrade one of the other technologies, how would we show that in quickstarter's version? Say, if the version is 2.2 (after spring-boot) and we upgrade the JDK from 11 to 12, would the quickstarter's version remain the same (-> 2.2)?
Same issue might come up if the quickstarter's version reflects the versions of multiple major technologies / frameworks, and we still have to change something else in the quickstarter. Then, the version would still remain unchanged? Which was already pointed out above...

metmajer: Hey @georgfedermann! I agree that a version for each quickstarter may come in handy, but let's keep it simple for now. I suggest we use the framework version in the `version` field and describe any other good-to-know versions in prose under `description`. Interested in adding to this PR?

metmajer:
@michaelsauter what do you think about putting the metadata into a dedicated metadata.yml? Frankly, those data are not release manager specific and could even power what the provisioning app displays for a technology. Even if that's ahead of what we want to achieve anytime soon, I'd like to put this data into a proper location. Thoughts?

michaelsauter:
@metmajer I'd be in favour of a `metadata.yml`. 

I've read the comments by @georgfedermann and like it so far. The suggestion also matches our current approach: ods-quickstarters is tagged with the version of ODS. So right now we care about the "quickstarter version", not the "tech version".

metmajer:
@michaelsauter @georgfedermann has raised a couple of good questions. To summarize what you say, all quickstarters have the version of ODS, correct?

georgfedermann:
From where I am standing, if I get the argument right, the approach that all quickstarters share one version, which would be the ods version, does not appear feasible. It looks like chaining things together that can change independently from each other.
If the ods version changes, so would the versions of all quickstarters, no matter if a specific quickstarter did change or not. This would generate a lot of noise for quickstarter users, because most of the version changes on the quickstarters would relate to changes in ods (or in other quickstarters), not in the quickstarter they are interested in itself. With each version change the quickstarter users will have to check whether something changed or not in the quickstarters they are interested in.
And the other way round, if we make a change to a quickstarter and need to bump its version, which is bound to the ods version, we would have to bumb the ods version to reflect the change to the quickstarter, right?
Also, if the quickstarters' versions are driven by the ods releases, then as a user I would expect to find information about quickstarter changes together with the ods release notes, right? While on the other hand, as I user I would expect quickstarters to have release notes on their own.

The same arguments  would apply if all quickstarters shared one version which is not bound to the ods version. We would still chain quickstarters together which can change independently from each other.

Would it be technically feasible to let each quickstarter manage its own version which would be incremented to reflect each change that was applied to the specific quickstarter, with an updated description reflecting the current setup of the quickstarter, maybe including release notes describing what changed?

metmajer:
I see the same issues. Looking at this through the eyes of the end-user, making a selection based on the ODS version doesn't make sense. I would give the framework version a clear preference (under the assumption we have to choose one version).

michaelsauter:
@metmajer @georgfedermann Maybe to give some perspective on this:

Using the ODS version as the quickstarter version is IMO the easiest approach to versioning. It basically says that you can use a quickstarter v3 if you have ODS v3 installed.

If you have separate versions, then you also need to advertise compatibility, which offers a new can of worms. To avoid this, quickstarter version = ODS version.

Yes this might raise the version of the quickstarter even if nothing changed. That is not ideal. However, doing a half-yearly release of ODS, one would expect some dependency update in the quickstarter anyway.

Now, this whole concept applies so far to the "official quickstarters" located in this repo. This repo cannot grow forever. Eventually quickstarters must be located in other repos because we cannot maintain all languages / frameworks.

Given that, we should start to advertise / check compatibility, but I believe first a concept would need to be developed for this. Therefore, for now a pragmatic approach is to use "quickstarter version = ODS version" until we have that concept.

And that I think is in line with @georgfedermann initial suggestion to have a version of the quickstarter, not of the languages / frameworks used within.

georgfedermann:
@michaelsauter ok, I can see your point. So, for the time being anyway, quickstarter releases are actually bound to ods releases and thus a common versioning scheme totally makes sense.

Maybe, at some point, we can have an exchange on how a new versioning system / compatibility check could work.

michaelsauter:
@georgfedermann cool. And please don't get me wrong: decoupling makes sense for me going forward. but we need to be able to handle it first! Maybe something for the v4 timeline (end of 2020)

tbugfinder:
Reading the initial request #75 it's about listing components and versions of a project (?) for the DSD. 
A project forked from a quickstarter might list the quickstarter and it's release assuming that it's not changed within the project.
However the quickstarter code can be changed by using a  patched version of component X within a project or by adding additional software. This creates a new lifecycle and would extend the listed Software components and versions for the DSD. This isn't part of the discussion yet, as per my understanding.

metmajer:
Ok, so these are the next steps:

1) `version` points to the ODS version. Versions of used technologies, starting with the framework, are listed in `description`
2) `metadata.yml` to contain re-usable metadata as it's not exclusive to the release manager

@georgfedermann want to adjust this PR?

georgfedermann:
> @georgfedermann want to adjust this PR?
@metmajer 👍

metmajer:
Closed via #143.

PR #132, opened 2020-01-23
using curl start.spring.io instead of spring init cli

### LINKED ISSUE #131: be-java-springboot spring init might fail if being under special corporate settings
**Describe the bug**
Current provisioning pipeline definition for `be-java-springboot` creates boilerplate's skeleton based on `spring init` (installed via `sdkman`).
This might not work in special network configuration conditions (e.g.: spring CLI might not properly load proxy environment variables or spring CLI HTTP requests to start.spring.io endpoint might be blocked).

**To Reproduce**
1. Go to prov-app
2. Provision `be-java-springboot`
3. Provisioning job might fail when running under special corporate network conditions

**Expected behavior**
The project skeleton creation must work - always

 - OpenShift: any
 - OpenDevStack 2.x, master

```bash
[Pipeline] }
[Pipeline] // stage
[Pipeline] stage
[Pipeline] { (Build spring project)
[Pipeline] sh
[newproject-cd-ods-quickstarter-be-java-springboot-be-java-springboot-2x] Running shell script
+ /home/jenkins/.sdkman/candidates/springboot/current/bin/spring init --build=gradle --java-version=1.8 --groupId=newproject --artifactId=be-java-springboot --package-name=com.bix_digital.newproject.be_java_springboot --packaging=jar --dependencies=web,jersey,data-jpa,h2,lombok,data-rest,restdocs,security be-java-springboot
Picked up JAVA_TOOL_OPTIONS: -XX:+UnlockExperimentalVMOptions -Dsun.zip.disableMemoryMapping=true
Using service at https://start.spring.io
Failed to retrieve metadata from service at 'https://start.spring.io' (Connection reset)
```

**Additional context**
This can be solved by using cURL CLI requests to start.spring.io

gerardcl:
fixes #131

### DIFF
--- a/be-java-springboot/Jenkinsfile
+++ b/be-java-springboot/Jenkinsfile
@@ -39,21 +39,23 @@ podTemplate(
           stage('Checkout quickstarter') {
             checkout scm
           }
-          stage('Install spring') {
-              sh 'curl -s "https://get.sdkman.io" | bash'
-              sh "/bin/bash -lc 'source /home/jenkins/.sdkman/bin/sdkman-init.sh && sdk install springboot ${springCliVersion} && spring --version'"
-          }
           stage('Build spring project') {
-            // TODO: Check if spring is in the PATH or think about alternative path
-            sh "/home/jenkins/.sdkman/candidates/springboot/current/bin/spring init \
-                --build=gradle \
-                --java-version='1.8' \
-                --groupId=${projectId} \
-                --artifactId=${componentId} \
-                --package-name='${packageName}' \
-                --packaging=jar \
-                --dependencies='web,jersey,data-jpa,h2,lombok,data-rest,restdocs,security' \
-                  ${componentId}"
+            sh "mkdir -p ${componentId}"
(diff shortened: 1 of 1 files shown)


michaelsauter:
I was wondering how that has worked previously. Because to my knowledge, we used spring init before, and somehow that must have worked. So - some digging shows that we have lost a wrapper script in the migration :(  See https://github.com/opendevstack/ods-project-quickstarters/pull/59. Maybe we should bring that back for v3?

gerardcl:
right! the "fun fact" here is that I was manually testing loading the environment variables and also through JAVA_OPTS/JAVA_TOOL_OPTIONS and that was not successful neither, that is why I decided to use the Spring Initializer official API (via cURL command you are provided if you manually configure the spring init through https://start.spring.io/ ). So, at the end, I like this option more since we do not need to download either sdkman nor install spring cli. But! I am fine if it is decided to restore the wrapper option (but that must be tested/validated in the same conditions)

PR #146, opened 2020-02-10
remove metadata structure from metadata.yml files

georgfedermann:
flattened metadata structure


metmajer:
Thanks @georgfedermann!

PR #153, opened 2020-03-07
add gunicorn and restructure test suite for be-python-flask 2.x

gerardcl:
related to opendevstack/ods-quickstarters/pull/152


be-python-flask/files/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.7.0-slim
+FROM python:3.6.10-slim
```
henrjk: What is the reason to stay on 3.6?

be-python-flask/files/tests/requirements.txt
```diff
@@ -0,0 +1,5 @@
+flask==1.1.1
+mypy==0.760
+flake8==3.7.9
+pytest==5.3.2
+pytest-cov==2.8.1
```
henrjk: pytest-cov 2.8.1 requires coverage>=4.4. When using coverage 5.x (in another context) I ran into an issue as the base image python did not compile the sqlite module (https://github.com/nedbat/coveragepy/issues/887)
Perhaps we should specify coverage==5.0.3 here and make sure we have a  python supporting sqlite or pin an earlier version.

common/jenkins-slaves/python/Dockerfile.rhel7
```diff
@@ -30,7 +31,7 @@ RUN curl "https://bootstrap.pypa.io/get-pip.py" -o "get-pip.py" \
 # Upgrade PIP
 RUN pip3 install --upgrade pip \
     && pip3 -V \
-    && pip3 install virtualenv pycodestyle
+    && pip3 install virtualenv==${VENV_VERSION}
```
henrjk: Are there reasons this is not using Pythons standard library venv module (https://docs.python.org/3/library/venv.html)?

be-python-flask/Jenkinsfile.template
```diff
@@ -30,49 +30,61 @@ odsPipeline(
...
 
+def stageTestSuite(def context, def nexusHostWithBasicAuth, def nexusHostWithoutScheme) {
+  String testLocation = 'build/test-results/test'
+  String coverageLocation = 'build/test-results/coverage'
+
+  stage('Prepare Test Suite') {
+    sh """
+      virtualenv testsuite
+      . ./testsuite/bin/activate
+      pip install -i ${nexusHostWithBasicAuth}/repository/pypi-all/simple --trusted-host ${nexusHostWithoutScheme} -r tests/requirements.txt
```
henrjk: If the user starts adding requirements these may not be compatible. However pip install will not return with a non-zero exit code in these cases. See https://github.com/pypa/pip/issues/6969.
I think adding a pip check after pip installs should do the trick.

be-python-flask/Jenkinsfile.template
```diff
@@ -30,49 +30,61 @@ odsPipeline(
...
+      mkdir -p ${testLocation}
+      mkdir -p ${coverageLocation}
+    """
+  }
+
+  stage('Lint') {
```
henrjk: As the application grows installing the dependencies for testing will take more time. Linting does not typically require that these dependencies are installed. An idea is to move linting in earlier stage so builds fail faster and perhaps even adding the linter in the jenkins slave image.

common/jenkins-slaves/python/Dockerfile.rhel7
gerardcl: I have checked if it is better to use venv instead of virutalenv and here you can find strong reasons to keep using virutalenv --> https://virtualenv.pypa.io/en/latest/

be-python-flask/Jenkinsfile.template
gerardcl: pip check in place! also in the Dockerfile :)

michaelsauter:
This looks like a feature to me so I wouldn't put this into 2.x?

Apart from that, the direction looks great - awesome work!

gerardcl:
hi @michaelsauter! thanks :)

I wanted to provide this already for ODS2 but was not in time indeed 

In this PR there are many bugfixes not only enhancements (new feature) that would be nice to have on ODS2 too.

What I can do is to cherry-pick and provide specific fixes only targedted for ODS2:
- packages versioning
- WSGI server
- python versions alignment between slave and quickstarter

is that better?

michaelsauter:
@gerardcl I would keep v2 as is. Unless the current quickstarter does not work at all (and I believe it to work at least somewhat, right?), I would not put your effort in there. Your are saying it yourself: "... that would be nice to have" ... I would not spend your energy there :)

gerardcl:
agree! then I will wait @henrjk  and/others approval and make the real PR go forward 

indeed the comments, except yours, should have gone to the master PR lol

yay!!

common/jenkins-slaves/python/Dockerfile.rhel7
henrjk: I would still vote to use the builtin module unless we actually run into limitations.

gerardcl: well, we want to enable users with other python versions for example, isn't that enough reason?

gerardcl: I see a point on being leaner and also that there is still the chance that if a user needs virutalenv he/she can install it via venv hehe, ok!!! but what abouth offering both venv and virutalenv from the beginning? :)

henrjk: I am not convinced, but I also did not see a problem.

gerardcl:
closing this ticket as agreed! :) 
related to /pull/152

PR #157, opened 2020-03-14
Add env var releaseStatusJiraIssueKey for e2e-spock-geb

metmajer:
This PR adds an env variable `releaseStatusJiraIssueKey` to the Releases Manager's Jenkins pipeline strategy. This field will contain the key to a Jira issue Release Manager needs to update about pipeline success and failure. This is necessary so that our Jira workflows can transition to a successor state.

We chose to use this approach over a convention-based approach, e.g. have the Release Manager reference a Jira issue by a label or similar, to avoid ambiguity.


metmajer:
@clemensutschig correct, and has been aligned with Fadi on Friday

Issue #175, opened 2020-03-21
MRO / Quality Release tracking issues for quickstarters


clemensutschig:
1. Each quickstarter (except airflow, r.shiny, jupyther, docker plain and release manager) has Sonarqube enabled (that implies working sonar-project.properties)

```
https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/vars/stageScanForSonarqube.groovy
```

and output is verified in SQ & jenkins log (e.g. Node.js currently fails with)

```
INFO: Load project repositories
INFO: Load project repositories (done) | time=6ms
ERROR: You are using version of TypeScript 2.9.2 which is not supported; supported versions >=3.2.1
ERROR: If it's not possible to upgrade version of TypeScript used by the project, consider installing supported TypeScript version just for the time of analysis
ERROR: Failure during analysis, Node.js command to start eslint-bridge was: 
````
 
1. Each quickstarter (except airflow, r.shiny, jupyther, docker plain and release manager) dumps the xUnit results into the right folder for shared lib to pick it up and stash it for the mro

```
https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/OdsPipeline.groovy#L164
```

1. Each quickstarter (except airflow, r.shiny, jupyther, , docker plain and release manager) calls junit to notify jenkins of the xUnit results

```
      junit 'build/test-results/test/*.xml'
```

1. Other related issues:
- related to #165 
- related to #166 
- related to #167

clemensutschig:
@metmajer fyi.

gerardcl:
hi! airflow now also has sonarqube (#174), I will provide jUnit fix
also solved OOM Killed issue in ds-rshiny ⚙️ -> #177 
seems fe-react mig

clemensutschig:
There is one general Topic we have to Talk about. Junit. We could do this centrally in the Shared lib.. where we collect results... or in every quickstarters build Phase..  thoughts??

metmajer:
Preferably into a single location, so there’s one less thing to care about.

metmajer:
@stitakis Here is the link to the junit plugin: https://jenkins.io/doc/pipeline/steps/junit/. You can use the `allowEmptyResults` attribute safely inside the ODS shared lib. MRO addresses missing test results if they are not needed but provided.

clemensutschig:
[...]

PR #178, opened 2020-03-22
adds sq to flask, adds junit to scala

stitakis:


stitakis:
This partially fixes #175:
- it adds sonarqube to flask qs
- it adds junit and copy test reports to standard test-report folder

metmajer:
@stitakis why don't we move the invocation of `junit` entirely into the ods shared library (to the location where the xUnit reports are stashed so we don't need to replicate this in every QS?

stitakis:
@metmajer yes, it will be cleaner/better and agree to proceed with your proposal once we fixed all QS open issues (or in parallel if someone picks this up in the meanwhile)

be-python-flask/Jenkinsfile.template
```diff
@@ -20,6 +20,7 @@ odsPipeline(
   ]
 ) { context ->
   stageBuild(context)
+  stageScanForSonarqube(context)
```
gerardcl: hi @stitakis , please, do not load a second time the `stageScanForSonarqube`, it was already existing in https://github.com/opendevstack/ods-quickstarters/blob/7e5f08e8dce7371fab79129d2429c86c5c6d7dee/be-python-flask/Jenkinsfile.template#L47 

the new implementation with the new test suite already in master (not added in 2.x due to breaking changes from the slave) has it more explicit.

be-scala-akka/Jenkinsfile.template
```diff
@@ -29,6 +29,14 @@ def stageBuild(def context) {
...
     withEnv(["TAGVERSION=${context.tagversion}"]){
       sh "sbt test"
+
+      sh 'pwd'
+      sh 'ls -la target/test-reports'
+      junit 'target/test-reports/**/*.xml'
+      sh 'mkdir -p build/test-results/test'
+      sh 'cp target/test-reports/*.xml build/test-results/test/'
+      sh 'ls -la build/test-results/test'
+
```
gerardcl: @stitakis please cleanup the code of `pwd`and `ls` commands. If you want verbosity you could add the respective flag in the `cp` command.

be-python-flask/Jenkinsfile.template
```diff
@@ -20,6 +20,7 @@ odsPipeline(
   ]
 ) { context ->
   stageBuild(context)
```
gerardcl: hi @stitakis, just in case, in this stage there is also already a call to `junit`, see: https://github.com/opendevstack/ods-quickstarters/blob/7e5f08e8dce7371fab79129d2429c86c5c6d7dee/be-python-flask/Jenkinsfile.template#L46

be-scala-akka/Jenkinsfile.template
clemensutschig: Why dont you use test Location??

be-python-flask/Jenkinsfile.template
stitakis: ok... I didn't expected the sq stage within stageBuild... anyway I reverted this change. So no change in flask qs

gerardcl: please, instead of adding a new commit for removing an issue added on previous commit, just rebert the commit itself, so no need to keep this in the history.

be-scala-akka/Jenkinsfile.template
```diff
@@ -29,6 +29,12 @@ def stageBuild(def context) {
   stage('Build and Unit Test') {
     withEnv(["TAGVERSION=${context.tagversion}"]){
       sh "sbt test"
+
+      junit 'target/test-reports/**/*.xml'
+
+      sh 'mkdir -p build/test-results/test/'
+      sh 'cp target/test-reports/*.xml build/test-results/test/'
```
gerardcl: I would suggest `cp -v`, so to show what is copied.

PR #192, opened 2020-03-23
ds-ml-service junit migration

gerardcl:


gerardcl:
@hugowschneider we redefine the folder here:
https://github.com/opendevstack/ods-quickstarters/blob/5ae2375226881f9f4b4eb89ad3decc6dc362a417/ds-ml-service/Jenkinsfile.template#L24

PR #204, opened 2020-03-23
Fix/nodejs testreport

ManuelFeller:


be-typescript-express/init.sh
```diff
@@ -24,7 +24,9 @@ cd $COMPONENT
 
 echo "generate project"
 yo --no-insight node-express-typescript
-
+npm install typescript@3.2.1
```
cschweikert: also with --save-dev ?

PR #215, opened 2020-03-25
Set value of completionDeadlineSeconds unquoted

### LINKED ISSUE #214: Template specifies int value as string
**Describe the bug**
https://github.com/opendevstack/ods-quickstarters/blob/6316170eabe854a941af07456307f205db4cf27d/common/ocp-config/component-environment/component-template.yml#L40-L44 defines a parameter for a field which is of type int. https://docs.openshift.com/container-platform/3.11/dev_guide/templates.html#writing-parameters explains how the param needs to be defined if it should be entered unquoted.

For some reason, when the BC is created, the string value is accepted and transformed to an int. But when it is updated (via `oc apply`) then it fails with:
```
Error from server: error when applying patch:
...
for: "STDIN": unrecognized type: int64
```

**To Reproduce**
Run `tailor apply` twice on the template.

**Expected behavior**
No drift should be detected, and no error should happen.

**Affected version:**
 - OpenShift: all
 - OpenDevStack: unsure, potentially only 2.x (as this looks related to the Tailor update to 0.13.1)

michaelsauter:
See
https://docs.openshift.com/container-platform/3.11/dev_guide/templates.html#writing-parameters.

Fixes #214.

### DIFF
--- a/common/ocp-config/component-environment/component-template.yml
+++ b/common/ocp-config/component-environment/component-template.yml
@@ -41,7 +41,6 @@ parameters:
     displayName: Completion Deadline Seconds
     description: how long the docker build of the component can last before it is canceled
     value: "1800"
-    required: true
 objects:
   - apiVersion: v1
     kind: Service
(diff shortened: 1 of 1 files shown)


michaelsauter:
Will merge after `v2.1` tag.

PR #218, opened 2020-03-26
change base image to ubi-minimal for be-scala-play

oalyman:


be-scala-play/files/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM openjdk:8-jre-alpine
+FROM adoptopenjdk/openjdk8:ubi-minimal-jre
```
michaelsauter: Ha, I thought the UBI's would come from RedHat's namespace. Does this mean that this image is based on an UBI base image from Redhat?

oalyman: Yes.

https://github.com/AdoptOpenJDK/openjdk-docker/blob/3b28dd3fb7230f208ed49da507432a01741b1259/11/jre/ubi-minimal/Dockerfile.hotspot.releases.full#L20

PR #219, opened 2020-03-30
updates spring boot dependency versions 

stitakis:
of asciidoctor, spring boot and spring restdocs to latest available


stitakis:
@renedupont I changed the wrong file. I've modified `build-4.10.build` which is the one that support legacy gradle version 4.1
I'll cancel this PR

PR #222, opened 2020-04-03
Use new shared quickstarter library for Go quickstarter

michaelsauter:
Based on work in https://github.com/opendevstack/ods-jenkins-shared-library/pull/230.


michaelsauter:
Adjusted after updates to #230.

Issue #213, opened 2020-03-25
Quickstarters need to generate code coverage (and report to SQ)


clemensutschig:
**Describe the bug**
Quickstarters lack consistent code coverage reporting, some have (jacoco) others don't. 

This is a must fix for ODS 3

This is a follow up from #175 

exit criteria: all qs of type ods / ods-test have the right code coverage show  up in SQ

michaelsauter:
Status:

- [ ] airflow-cluster
- [x] ~be-fe-mono-repo-plain~
- [ ] be-gateway-nginx
- [x] be-golang-plain
- [x] be-java-springboot
- [ ] be-python-flask
- [ ] be-scala-play
- [ ] be-typescript-express
- [x] ~docker-plain~
- [ ] ds-jupyter-notebook
- [ ] ds-ml-service
- [ ] ds-rshiny
- [ ] e2e-cypress
- [ ] e2e-spock-geb
- [ ] fe-angular
- [ ] fe-ionic
- [ ] ods-document-gen-svc
- [ ] ods-provisioning-app
- [x] ~release-manager~

QS with no owners / about to be removed:
- [ ] be-scala-akka
- [ ] fe-react
- [ ] fe-vue

gerardcl:
-for `ds-jupyter-notebook` and `ds-rshiny` I think they do not require it based on the type of service they are
-for `airflow-cluster` we are in conversations to move it to extra-quickstarters
-regarding `be-python-flask`and `ds-ml-service`: AFAIK I already implemented that, isn't it working anymore?

michaelsauter:
@gerardcl I simply did not look into Python / ML. Updated the list now, thanks for the comments!

bljubisic:
For fe-ionic there is a branch Fix-Ionic-Unit-Test-Reporting which is a fix for Ionic.

michaelsauter:
Closing this as all quickstarters have code coverage configured now!

PR #226, opened 2020-04-08
Configure JaCoCo to generate XML reports

### LINKED ISSUE #225: be-java-springboot generates incompatible JaCoCo configuration
**Describe the bug**
The Jacoco config worked with the SonarQube Java plugin v5 and lower. v6 of the plugin removed JaCoCo support in favour of installing a new JaCoCo plugin, which only works with an XML report.

**Affected version**
 - OpenDevStack: master, 2.x

michaelsauter:
Fixes #225.

@clemensutschig I would actually propose to apply this fix for 2.x as well, since code coverage is broken there since the update of the Java plugin to 6.2. Do you agree?


michaelsauter:
Implements #213 for the Java SpringBoot quickstarter.

PR #233, opened 2020-04-16
MRO / monorepo quickstarter fixes

clemensutschig:
quickstarter based on opendevstack/ods-jenkins-shared-library#242

2 buildconfigs / imagestreams
1 deployment / 2 containers 

@martsec  / @metmajer  :)


be-fe-mono-repo-plain/README.md
```diff
@@ -0,0 +1,5 @@
+# Backend - Go (be-golang)
```
gerardcl, line 1: this should be changed right?

gerardcl: I would suggest to assure it is also explained further so people know all options available now regarding monorepo structures:
1- different components in one repo and one deploymentconfig (1podNcontainers)
2- different components in one repo and different deploymentconfigs (Npods)
3- mix of 1 and 2

PR #235, opened 2020-04-16
Convert Python quickstarter

michaelsauter:


be-python-flask/Jenkinsfile
```diff
@@ -1,77 +1,18 @@
...
-    ) {
-        node(podLabel) {
-          stage('Checkout quickstarter') {
-            checkout scm
-          }
-          stage("initialize target folder ${componentId}") {
-              sh "mkdir -p ${componentId}"
-          }
-          stage('copy files from quickstart to generated project') {
-              sh "cp -rv ${quickstarterBasepath}/README.md ${quickstarterBasepath}/files"
```
gerardcl, line 45: approved!

just one question: why are not we providing an initial readme file to the provisioned quickstarters as done here for example?

michaelsauter: hmm ... I guess up to the quickstarter owner? In general, this is probably a good idea :)

gerardcl: shall I make a proposal fix on the ODS lib side then? so to have also a cp of the readme into the files/target folder?

michaelsauter: @gerardcl I'd rather leave this to the quickstarter author. If they place a `README.md` into `files`, great. Otherwise, it will not be there. I think a `README.md` in the "quickstarter directory" can (and should?) serve a different purpose.

PR #237, opened 2020-04-18
Fix/ds ml service mro mono repo

### LINKED ISSUE #231: Make ds-ml-quickstarter work with mono-repo MRO implementation
follow up from opendevstack/ods-jenkins-shared-library#242 and opendevstack/ods-mro-jenkins-shared-library#188

1. use the new stages :) to remove all the clutter, that is now standard 
1. for ds-ml quickstarter to work with the MRO - the labels of all components - that need promotion - have to use the standard pattern `app: '${PROJECT}-${COMPONENT}'` 

as of now the quickstarter sets the component to be `${COMPONENT}`-training/prediction-service, which tailor does not know if in the MRO context (it searches for `${COMPONENT}` being the repo.id.

Second thought: we have to talk about what promote to production means ....
assuming you'd call the component environment on the first attempt - as is -> without `-training..` for the prediction service we would pick only that one up - and transport ..

BUT this highly depends on if you want to retrain on Q and P ... 

NB: I tested the mro with the standard `app` selector, and this works now

gerardcl:
fixes #231


gerardcl:
@clemensutschig could you also test/verify it?

clemensutschig:
> @clemensutschig could you also test/verify it? or can you indicate me where did you test the others (considering provisioning from scratch, and not modifying existing jenkinsfile of a provisioned one) like the multi repo quickstarter?

I used project `odsst` (component: `machine`) on premise ... what you need to verify is that each artifact has the right `app` selector .. I see a bunch of those being still set to `-training/prediction` ..

gerardcl:
> > @clemensutschig could you also test/verify it? or can you indicate me where did you test the others (considering provisioning from scratch, and not modifying existing jenkinsfile of a provisioned one) like the multi repo quickstarter?
> 
> I used project `odsst` (component: `machine`) on premise ... what you need to verify is that each artifact has the right `app` selector .. I see a bunch of those being still set to `-training/prediction` ..

Hi Clemens! I did base/compare from the resources in `odsst` and:
- `metadata:labels:app` definitions match the component id only (this does not appear in the template because it is set via tailor when applying it), this is where tailor defines selectors/labels
- `spec:selector:app` definitions have the same values I found in the resources in `odsst` <- do I need anyway to define them as per deploymentconfig?

I will check that last statement but then that will not match the deployed resources in `odsst`

clemensutschig:
> > > @clemensutschig could you also test/verify it? or can you indicate me where did you test the others (considering provisioning from scratch, and not modifying existing jenkinsfile of a provisioned one) like the multi repo quickstarter?
> > 
> > 
> > I used project `odsst` (component: `machine`) on premise ... what you need to verify is that each artifact has the right `app` selector .. I see a bunch of those being still set to `-training/prediction` ..
> 
> Hi Clemens! I did base/compare from the resources in `odsst` and:
> 
> * `metadata:labels:app` definitions match the component id only (this does not appear in the template because it is set via tailor when applying it), this is where tailor defines selectors/labels
> * `spec:selector:app` definitions have the same values I found in the resources in `odsst` <- do I need anyway to define them as per deploymentconfig?
> 
> I will check that last statement but then that will not match the deployed resources in `odsst`

check out odsst -> machine .. if similar, it'll work :)

ds-ml-service/openshift/component-template.yml
```diff
@@ -0,0 +1,336 @@
...
+          intervalSeconds: 1
+          maxSurge: 25%
+          maxUnavailable: 25%
+          timeoutSeconds: 600
+          updatePeriodSeconds: 1
+        type: Rolling
+      template:
+        metadata:
+          labels:
+            app: '${PROJECT}-${COMPONENT}-prediction-service'
```
clemensutschig, line 204: @gerardcl  :)

gerardcl:
That is also the same value in the resource in odsst.
Anyway, it still does not 100% work yet (at least on BIX cluster) when triggering the deployment, I amb using the same shared lib branch and so on though.
I will continue late today and tomorrow, will come back to you :)

clemensutschig:
@gerardcl .. You need to take also the mro pr... i am working on getting r and jupyther to fly as well.. here the authproxy is the challenge.. ;-)

gerardcl:
ready to go :)
I was missing using the right branch for both libraries and then noticed I also was missing the definition of the buildArgs.

DONE :)

gerardcl:
@clemensutschig regarding the oauth-proxy we could have our fork since we could have the customizations Florian made by his own to assure there is never a timeout if no usage of the dashboards. Just keep that as an option for your issue if not that might come in the future anyway.

PR #260, opened 2020-04-27
WIP: Add gcc/g++ to support builds with CGO_ENABLED=1

### LINKED ISSUE #230: Add gcc/g++ to support builds with CGO_ENABLED=1
**Describe the bug**
Golang builds can either yield a static, fully self-contained binary, or be dynamically-linked against a C library. Currently, when building the latter using `CGO_ENABLED=1`, the build fails because of a missing `gcc`.

**To Reproduce**
`CGO_ENABLED=1 go build`

**Expected behavior**
The build succeeds.

 - OpenShift: [e.g. 3.9, 3.11]
 - OpenDevStack 2.x

**Proposed solution**
Add `gcc` and `gcc-c++` into the Golang build slave. The presence of both further allows developers to import C/C++ files into Golang. For this to work, in the Dockerfile, add:

```
RUN yum install -y gcc gcc-c++
```

metmajer:
Closes #230

### DIFF
--- a/common/jenkins-slaves/golang/Dockerfile.rhel7
+++ b/common/jenkins-slaves/golang/Dockerfile.rhel7
@@ -4,7 +4,7 @@ LABEL maintainer="Michael Sauter <michael.sauter@boehringer-ingelheim.com>"
 
 ENV GO_VERSION 1.14.2
 
-RUN yum install -y gcc gcc-c++ make
+RUN yum install -y gcc gcc-c++
 
 RUN cd /tmp && \
     curl -LO https://storage.googleapis.com/golang/go$GO_VERSION.linux-amd64.tar.gz && \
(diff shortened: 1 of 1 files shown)


metmajer:
Also adds `make` as it goes well with Go.

metmajer:
@michaelsauter that would make most sense. Let me remove it from here again.

michaelsauter:
@metmajer Perfect, please open an issue in ods-core to add it there.

metmajer:
@michaelsauter cheers, will do soon

michaelsauter:
@metmajer Do you want to merge this PR? As it is still labeled with `WIP`? Seems ready for me ...

tbugfinder:
Does the gcc/g++ version matter? How would a user get to a later version if needed?

michaelsauter:
@tbugfinder If a later version is needed, users would need to build their own slave (we should have a guide how to do that - it is not that difficult) or contribute a new version to ODS.

metmajer:
@michaelsauter @clemensutschig so the open question here has been the following: making gcc and g++ available is one thing. The dependency of `CGO_ENABLED=1` on a C library is another one. The latter needs to exist in an ODS component that wants to make use of this feature. Otherwise, users run into a "file not found" error. I needed to test-drive the approach first in a Golang ODS component before coming to a conclusion, hence the `WIP`.

PR #276, opened 2020-05-06
Allow configuration of BB project

michaelsauter:
Follow-up from https://github.com/opendevstack/ods-core/pull/538.


michaelsauter:
@clemensutschig Those files do not reference the Bitbucket `opendevstack` project. They only reference images, which are in the OpenShift `ods` namespace.

Issue #277, opened 2020-05-06
Add C library to Golang quickstarter to support CGO_ENABLED=1


metmajer:
**Is your feature request related to a problem? Please describe.**
After the recent change in https://github.com/opendevstack/ods-quickstarters/pull/260, we want to enable users to effectively make use of `CGO_ENABLED=1` by including a C library into the Golang quickstarter.

**Describe the solution you'd like**
Add a C library that is supported by the base image.

jorge-romero:
Currently there was a change done in the test application were we used a prebuilt library as it is not provided to Alpine. I can be installed easily with the following script:

# Here we install GNU libc (aka glibc) and set C.UTF-8 locale as default.

RUN ALPINE_GLIBC_BASE_URL="https://github.com/sgerrand/alpine-pkg-glibc/releases/download" && \
    ALPINE_GLIBC_PACKAGE_VERSION="2.31-r0" && \
    ALPINE_GLIBC_BASE_PACKAGE_FILENAME="glibc-$ALPINE_GLIBC_PACKAGE_VERSION.apk" && \
    ALPINE_GLIBC_BIN_PACKAGE_FILENAME="glibc-bin-$ALPINE_GLIBC_PACKAGE_VERSION.apk" && \
    ALPINE_GLIBC_I18N_PACKAGE_FILENAME="glibc-i18n-$ALPINE_GLIBC_PACKAGE_VERSION.apk" && \
    apk add --no-cache --virtual=.build-dependencies wget ca-certificates && \
    echo \
        "-----BEGIN PUBLIC KEY-----\
        MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEApZ2u1KJKUu/fW4A25y9m\
        y70AGEa/J3Wi5ibNVGNn1gT1r0VfgeWd0pUybS4UmcHdiNzxJPgoWQhV2SSW1JYu\
        tOqKZF5QSN6X937PTUpNBjUvLtTQ1ve1fp39uf/lEXPpFpOPL88LKnDBgbh7wkCp\
        m2KzLVGChf83MS0ShL6G9EQIAUxLm99VpgRjwqTQ/KfzGtpke1wqws4au0Ab4qPY\
        KXvMLSPLUp7cfulWvhmZSegr5AdhNw5KNizPqCJT8ZrGvgHypXyiFvvAH5YRtSsc\
[...]

PR #281, opened 2020-05-08
Add glibc to the golang quick starter

jorge-romero:
This will solve issue #277 
It use the same aprox that it is used in the Alpine openjdk official containers.

Co-authored-by: Juan Antonio Farre <jafarre@viewnext.com>

### DIFF
--- a/be-golang-plain/files/docker/Dockerfile
+++ b/be-golang-plain/files/docker/Dockerfile
@@ -1,5 +1,26 @@
 FROM alpine
 
+# Set locale
+ENV LANG=C.UTF-8
+
+# Install glibc so that CGO can be enabled when using dynamic linking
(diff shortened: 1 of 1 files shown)


henrjk:
My first impulse is to not use alpine if one would like to use glibc as alpine is build on musl and can cause unexpected issues.

metmajer:
@michaelsauter @henrjk currently, CGO is unusable. Uncommenting would make it hard to overlook and good for the moment. For security reasons alone, we should soon consider to adding to more secure base images anyway and resolve the current issue in a better way then.

metmajer:
@jorge-romero can you comment out the change you have added as a first approach?

jorge-romero:
Just commented the lines that sets and install glibc

Issue #21, opened 2019-06-21
Update Ionic quickstarter to newest Angular version


cschweikert:
Ionic quickstarter still uses Angular 5. It should be updated to use at least version 8 like in fe-angular, if possible.

This update should also include the improvements of SonarQube support, which were introduced on fe-angular. See also https://github.com/opendevstack/ods-project-quickstarters/issues/212

bljubisic:
As a first step I would like to recommend to modify fe-ionic quickstarter to follow the same principle as fe-angular quickstarter. After that we can modify the quickstarter to use latest ionic update.

PR #282, opened 2020-05-10
Update ionic to version 5 #21

bljubisic:
Fixing the Angular version used in Ionic quickstarter.


bljubisic:
Yes, with Manu and Janet. We created a video for documentation for all FEs :)

michaelsauter:
@bljubisic Nice. What would be the best way to make this available? Maybe you can upload it e.g. on YouTube and link it from the docs somewhere?

ManuelFeller:
Regarding publishing the Video: we plan do collect feedback internally - and then (once we have blurred out the URL's that are for internal use only) upload it maybe. We could create an OpenDevStack Video Channel or use the Members personal ones... What do you prefer @michaelsauter ?

michaelsauter:
@ManuelFeller official one would be cool - get in touch with @rattermeyer if there is already one.

Issue #275, opened 2020-05-05
Build fails for yarn install on nodejs slave


metmajer:
**Describe the bug**
Builds on the nodejs slave are failing when using `yarn install` because yarn is not configured to use a corporate proxy. Builds do not fail when using `npm install` instead.

**To Reproduce**
Run `yarn install`in an ODS components in our enterprise environment.

**Affected version (please complete the following information):**
 - OpenDevStack 2.x

**Log Output (ensure to remove any confidential information like tokens, project names, etc.**
```
+ yarn install
yarn install v1.22.4
[1/4] Resolving packages...
[2/4] Fetching packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/iconv-lite/-/iconv-lite-0.4.13.tgz: unexpected end of file".
info If you think this is a bug, please open a bug report with the information provided in "/tmp/workspace/pltfmdev-cd/pltfmdev-cd-demo-app-front-end-master/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

**Additional context**
A proxy and nexus needs to be configured as described at https://www.jhipster.tech/configuring-a-corporate-proxy/.

metmajer:
FYI @clemensutschig

clemensutschig:
@metmajer Workaround.  Try: https://www.garysieling.com/blog/using-sonatype-nexus-yarn-npm-replacement/amp/

clemensutschig:
@ManuelFeller can you fix this..  ? Default setup should point to nexus

ManuelFeller:
Will check tomorrow

ManuelFeller:
Hi. Question: do we talk about the corporate proxy or an internal registry?

Since we have no setting for a proxy in the `npmrc` we deploy I am surprised to hear that that would work - or is it just unaffected because the nexus host is already set as source and does not need to go through the proxy?

Second: are we OK to change the secrets for this slave?
I am asking because I have trouble finding any location where we could save an auth-token like the one used in the `npmrc`. Right now it seems that yarn would require to get username and password in the passed URL via `yarn config` commands. So we need to either change the secrets for this one or add redundant new ones.

I can imagine to change our current file copy method for npm to a bunch of command line calls performing the actions, so that we have a unified way for both packet managers - instead of handling npm in one way and yarn in another...

clemensutschig:
[...]

PR #285, opened 2020-05-13
changes required to make yarn use the internal registry by default

ManuelFeller:
fix for #275 - according to my testing this makes sure the slave is defaulted to the internal registry for npm and yarn

### DIFF
--- a/common/jenkins-slaves/nodejs10-angular/Dockerfile.rhel7
+++ b/common/jenkins-slaves/nodejs10-angular/Dockerfile.rhel7
@@ -61,17 +61,13 @@ RUN curl --silent --location https://rpm.nodesource.com/setup_10.x | bash - && \
     yum install -y yarn && \
     yum clean all -y
 
-COPY npmrc $HOME/.npm-global/etc/npmrc
-COPY npmrc $HOME/.npmrc
-COPY yarnrc $HOME/.yarnrc
-
-RUN sed -i "s|NEXUS_HOST|$NEXUS_HOST|g" $HOME/.npm-global/etc/npmrc && \
-    sed -i "s|NEXUS_AUTH|$(echo -n $NEXUS_AUTH | base64)|g" $HOME/.npm-global/etc/npmrc && \
-    sed -i "s|NEXUS_HOST|$NEXUS_HOST|g" $HOME/.npmrc && \
-    sed -i "s|NEXUS_AUTH|$(echo -n $NEXUS_AUTH | base64)|g" $HOME/.npmrc && \
-    sed -i "s|NEXUS_HOST|$NEXUS_HOST|g" $HOME/.yarnrc && \
+RUN npm config set registry=$NEXUS_HOST/repository/npmjs/ && \
+    npm config set always-auth=true && \
+    npm config set _auth=$(echo -n $NEXUS_AUTH | base64) && \
+    npm config set email=no-reply@opendevstack.org && \
     npm config set ca=null && \
     npm config set strict-ssl=false && \
+    yarn config set registry $NEXUS_HOST/repository/npmjs/ -g && \
     npm install -g @angular/cli@8.0.1 --unsafe-perm=true --allow-root && \
     npm install -g cypress@3.3.1 --unsafe-perm=true --allow-root > /dev/null && \
     npm --version && \
--- a/common/jenkins-slaves/nodejs10-angular/Dockerfile.rhel7
+++ b/common/jenkins-slaves/nodejs10-angular/Dockerfile.rhel7
@@ -62,9 +62,14 @@ RUN curl --silent --location https://rpm.nodesource.com/setup_10.x | bash - && \
     yum clean all -y
 
 COPY npmrc $HOME/.npm-global/etc/npmrc
+COPY npmrc $HOME/.npmrc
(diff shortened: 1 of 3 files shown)


common/jenkins-slaves/nodejs10-angular/yarnrc
```diff
@@ -0,0 +1 @@
+registry "NEXUS_HOST/repository/npmjs//"
```
ManuelFeller, line 1: the // at the end is required to make sure one of them stays when the value is used

common/jenkins-slaves/nodejs10-angular/Dockerfile.rhel7
```diff
@@ -62,9 +62,14 @@ RUN curl --silent --location https://rpm.nodesource.com/setup_10.x | bash - && \
     yum clean all -y
 
 COPY npmrc $HOME/.npm-global/etc/npmrc
+COPY npmrc $HOME/.npmrc
+COPY yarnrc $HOME/.yarnrc
```
ManuelFeller, line 66: yarn requires the .npmrc and .yarnrc to be directly in the home directory - else it will not work as it should (according to my tests)

common/jenkins-slaves/nodejs10-angular/yarnrc
cschweikert: I just wanted to ask ;)

common/jenkins-slaves/nodejs10-angular/Dockerfile.rhel7
```diff
@@ -62,9 +62,14 @@ RUN curl --silent --location https://rpm.nodesource.com/setup_10.x | bash - && \
...
 
 RUN sed -i "s|NEXUS_HOST|$NEXUS_HOST|g" $HOME/.npm-global/etc/npmrc && \
     sed -i "s|NEXUS_AUTH|$(echo -n $NEXUS_AUTH | base64)|g" $HOME/.npm-global/etc/npmrc && \
+    sed -i "s|NEXUS_HOST|$NEXUS_HOST|g" $HOME/.npmrc && \
+    sed -i "s|NEXUS_AUTH|$(echo -n $NEXUS_AUTH | base64)|g" $HOME/.npmrc && \
```
gerardcl, line 71: proposal: instead of having two times this file/creation we could use symbolic linking:
```
ln -s $HOME/.npm-global/etc/npmrc $HOME/.npmrc
```

common/jenkins-slaves/nodejs10-angular/Dockerfile.rhel7
gerardcl: if we do the symbolic linking suggestion above we could then avoid the second `COPY npmrc $HOME/.npmrc `

michaelsauter: Given that Yarn expects the files in `$HOME` - is there a way to tell NPM to use that, instead of `$HOME/.npm-global/etc/npmrc`?

common/jenkins-slaves/nodejs10-angular/yarnrc
michaelsauter: Maybe add a comment why it is like that for future puzzled readers? :)

common/jenkins-slaves/nodejs10-angular/Dockerfile.rhel7
ManuelFeller: I will check on this; will update you soon

ManuelFeller: Given [this documentation](https://docs.npmjs.com/files/npmrc) we should be able to remove the global one since it is resolving project specific, then user specific and then the two more global ones.

Also it seems to be possible that we can use environment variables directly in there - so I will test if yarn can do the same - then we can get rid of some or maybe all "sed magic" for that

ManuelFeller:
It seems that .yarnrc can not handle any environment variables in there, but by passing them via the command line config tool way we can utilize it.

And then (for clean up and unification) I changed all custom settings to be applied the same way (using the tools config commands instead of file-copy).

Worked nicely on a test-pod...

common/jenkins-slaves/nodejs10-angular/Dockerfile.rhel7
```diff
@@ -61,12 +61,13 @@ RUN curl --silent --location https://rpm.nodesource.com/setup_10.x | bash - && \
...
     yum clean all -y
 
-COPY npmrc $HOME/.npm-global/etc/npmrc
-
-RUN sed -i "s|NEXUS_HOST|$NEXUS_HOST|g" $HOME/.npm-global/etc/npmrc && \
-    sed -i "s|NEXUS_AUTH|$(echo -n $NEXUS_AUTH | base64)|g" $HOME/.npm-global/etc/npmrc && \
+RUN npm config set registry=$NEXUS_HOST/repository/npmjs/ && \
+    npm config set always-auth=true && \
+    npm config set _auth=$(echo -n $NEXUS_AUTH | base64) && \
+    npm config set email=no-reply@opendevstack.org && \
```
michaelsauter, line 67: Beautiful!

PR #297, opened 2020-05-22
Fix Jenkins slaves build config (add resource constraints)

michaelsauter:
* Add resource constraints
* Align NodeJS 12 slave template with other slaves
* Align Makefile targets with naming used in ods-core

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,76 +8,76 @@ NAMESPACE=ods
 
 # JENKINS
 ## Install or update Jenkins agent resources.
-jenkins: jenkins-apply-build jenkins-start-build
+jenkins: apply-jenkins-build start-jenkins-build
 .PHONY: jenkins
 
 ## Update OpenShift resources related to Jenkins agent images.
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -8,76 +8,76 @@ NAMESPACE=ods
...
 ## Update OpenShift resources related to Jenkins maven agent image.
-jenkins-apply-build-maven:
+apply-jenkins-build-maven:
 	cd common/jenkins-slaves/maven/ocp-config && tailor apply --namespace ${NAMESPACE}
-.PHONY: jenkins-apply-build-maven
+.PHONY: apply-jenkins-build-maven
 
 ## Update OpenShift resources related to Jenkins nodejs agent image.
-jenkins-apply-build-nodejs:
+apply-jenkins-build-nodejs10-angular:
```
gerardcl, line 34: don't we miss nodejs12? or not required yet?

michaelsauter: Well - I wasn't sure. It is not referenced anywhere so I decided to leave that like it is for now.

Issue #165, opened 2020-03-20
e2e cypress not compatible with mro for e2e testing


clemensutschig:
**Describe the bug**
Newly provisioned e2e cypress quickstarter - if added to the mro with type `ods-test` fails because mro expects three xmlUnit files

java.lang.RuntimeException: Error: unable to unstash JUnit XML reports for repo 'e2e-cypress' from stash 'acceptance-test-reports-junit-xml-e2e-cypress-40'.
	at phaseTest.getTestResults(phaseTest.groovy:109)
	at phaseTest.getAcceptanceTestResults(phaseTest.groovy:87)
	at phaseTest.call(phaseTest.groovy:43)
	at org.ods.util.MROPipelineUtil.prepareExecutePhaseForRepoNamedJob(MROPipelineUtil.groovy:285)
	at org.ods.util.PipelineUtil.executeBlockAndFailBuild(PipelineUtil.groovy:108)
	at org.ods.util.MROPipelineUtil.prepareExecutePhaseForRepoNamedJob(MROPipelineUtil.groovy:232)
	at ___cps.transform___(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at org.codehaus.groovy.reflection.CachedConstructor.invoke(CachedConstructor.java:83)
	at org.codehaus.groovy.reflection.CachedConstructor.doConstructorInvoke(CachedConstructor.java:77)
	at 

**Affected version (please complete the following information):**
 - OpenDevStack [2.x/master]

**Additional context**
https://github.com/opendevstack/ods-mro-jenkins-shared-library/blob/master/vars/phaseTest.groovy#L49-L57 

Junit reporter:
https://www.npmjs.com/package/cypress-junit-reporter

@metmajer  fyi

metmajer:
Thanks @clemensutschig. Should we one of a kind with the e2e-spock-geb QS.

michaelsauter:
@herrkoch Could you take a look at this?

herrkoch:
Hi @michaelsauter, I am happy to help, I noted it down and will take a look the issue as soon as I find some time!

[...]

PR #302, opened 2020-05-22
Fix/cypress #165 junit xml reports

herrkoch:
Quickstarter runs fine locally and on Jenkins; not yet tested with release manager (provisioning of the release manager failed in my playground)

### DIFF
--- a/e2e-cypress/Jenkinsfile.template
+++ b/e2e-cypress/Jenkinsfile.template
@@ -17,8 +17,7 @@ def stageTest(def context) {
   stage('Integration Test') {
     withEnv(["TAGVERSION=${context.tagversion}", "NEXUS_HOST=${context.nexusHost}","OPENSHIFT_PROJECT=${context.targetProject}","OPENSHIFT_APP_DOMAIN=${context.getOpenshiftApplicationDomain()}"]) {
       sh "npm install"
-      sh "npm run lint"
-      sh "npm run build"
+      sh "npm run all"
     }
   }
 }
--- a/e2e-cypress/Jenkinsfile.template
+++ b/e2e-cypress/Jenkinsfile.template
@@ -1,9 +1,9 @@
 // See https://www.opendevstack.org/ods-documentation/ for usage and customization.
 
-@Library('ods-jenkins-shared-library@@ods_git_ref@') _
+@Library('ods-jenkins-shared-library@3.x') _
 
 odsComponentPipeline(
-  imageStreamTag: '@ods_namespace@/jenkins-slave-nodejs10-angular:@ods_image_tag@',
+  imageStreamTag: 'ods/jenkins-slave-nodejs10-angular:3.x',
   branchToEnvironmentMapping: [
     'master': 'dev',
     // 'release/': 'test'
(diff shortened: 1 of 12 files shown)


clemensutschig:
@herrkoch whats the error you got for the rm?

herrkoch:
Hi @clemensutschig I get this error when provisioning the rm on my playground ("daniel" namespace on our OpenShift): `org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: Scripts not permitted to use method groovy.lang.GroovyObject getProperty java.lang.String (org.ods.quickstarter.Context.bitbucketUrl)`
I am only generating one JUnit xml file with the results of all Cypress e2e tests and store it at the expected location (as there is no differentiation between acceptance / integration / installation in the Cypress quickstarter). Should we introduce these 3 separate "stages" also in Cypress?

metmajer:
@herrkoch Hi Daniel and thanks for your contribution. The release manager expects three files, as otherwise, it cannot distinguish between these test types. Depending on the type, the test results need to go into different documents. Can you please fix that? Best you take a look at the spock/geb quickstarter and implement it as similarly as possible. Cheers!

herrkoch:
Thanks for the explanation @metmajer , I think I understand now what has to be done. My idea to solve this is to execute `cypress run` 3 times in a row, each time with a different config for `integrationFolder` and `mochaFile` for installation / acceptance / integration. First I'll follow up with one of the BI X colleagues on the release manager deployment issue.

metmajer:
Most welcome, @herrkoch. The issue you are describing above happens if you access a property of an object that does not exist. Are you using a version of ODS that does not have `org.ods.quickstarter.Context.bitbucketUrl`? Is it written in the correct way?

clemensutschig:
Context.bitbucketHost.. ;-) misleading jenkins error

herrkoch:
Hi @metmajer @clemensutschig I adapted the quickstarter to now execute e2e tests in 3 stages (acceptance + installation + integration), each one generating a JUnit XML result file that is being stashed by Jenkins. @michaelsauter helped me to fix the deployment of the release manager in my playground (daniel-cd) and achieved a successful run with the cypress quickstarter configured as dependency.

herrkoch:
It's worth noting that in the cypress quickstarter execution on Jenkins this happens as well:
`Stashing testResults from location: 'build/test-results/test'
[Pipeline] sh (Moving test results to system location: build/test-results/test)
+ mkdir -p build/test-results/test build/test-results/test
+ cp -rf 'build/test-results/test/*' build/test-results/test
+ true
cp: cannot stat 'build/test-results/test/*': No such file or directory
[Pipeline] sh (Counting test results in build/test-results/test)
+ wc -l
+ ls -la 'build/test-results/test/*.xml'
ls: cannot access build/test-results/test/*.xml: No such file or directory
[Pipeline] echo
No xUnit results found!!`
Not sure where this comes from but if I put the result files to `build/test-results/test` it will be stashed "automatically" but not picked up by the release manager. I had to include the same lines of code in the Jenkinsfile.template as for the spock/geb quickstarter to properly stash the results files so they are picked up by the release manager.

clemensutschig:
All good this Magic happens somewhere else.. no need to do this.  :-) Check
the spock qs.. there you see the stashing..

Daniel Koch <notifications@github.com> schrieb am Fr., 29. Mai 2020, 18:41:

e2e-cypress/Jenkinsfile.template
```diff
@@ -1,9 +1,9 @@
 // See https://www.opendevstack.org/ods-documentation/ for usage and customization.
 
-@Library('ods-jenkins-shared-library@@ods_git_ref@') _
+@Library('ods-jenkins-shared-library@3.x') _
 
 odsComponentPipeline(
-  imageStreamTag: '@ods_namespace@/jenkins-slave-nodejs10-angular:@ods_image_tag@',
+  imageStreamTag: 'ods/jenkins-slave-nodejs10-angular:3.x',
```
michaelsauter: You need to leave the variables in - that should always be replaced when the quickstarter is provisioned with the versions that the user chose.

herrkoch: Hi Michael, yes this was an accidental copy/paste change that I already fixed last Friday directly after my first commit (I guess that's why GitHub says "Outdated").

michaelsauter: Sorry, didn't realise. It showed that to me as part of the diff. Now if I go there it isn't there anymore ...

herrkoch: Yeah, I guess GitHub send you an email with a direct link to the changes of the first commit. They should change that and make it a link to the changes of the full PR. :-)

herrkoch:
Regarding the stashing: I copied that part over from the Jenkinsfile of the spock/geb quickstarter (with some path adjustments) and the stashing/unstashing worked fine in my test run with the release manager.

PR #306, opened 2020-05-27
fix path to coverage analysis data

cschweikert:
The path to the lcov file also needs the component id. The path is usually defined within the ``karma.conf.js`` under ``coverageIstanbulReporter``. ``karma.conf.js`` gets generated during quickstarter provisioning by the ``ng new`` command.

this fix also relates to #213

### DIFF
--- a/fe-angular/sonar-project.properties.template
+++ b/fe-angular/sonar-project.properties.template
@@ -20,4 +20,4 @@ sonar.coverage.exclusions=**/*.spec.ts
 sonar.exclusions=**/*.html,**/*.scss,**/*.json,**/*.ico,**/*.svg
 
 # Define coverage information inside the Angular project
-sonar.typescript.lcov.reportPaths=coverage/lcov.info
+sonar.javascript.lcov.reportPaths=coverage/@component_id@/lcov.info


michaelsauter:
Would it be possible to have the path at a fixed location as well? Having a dynamic part in the path makes me a bit uneasy. But if that's the default I'm, fine with it too.

Would we need the exact same fix for Ionic too?

cschweikert:
fe-ionic is a different situation: provision overwrites the generated ``karma.conf.js`` by ``fe-ionic/files/karma.conf.js``, which again uses the default path ``./coverage``. Should be all good here.

Since Angular version 6 the ``ng new $COMPONENT`` (from here https://github.com/opendevstack/ods-quickstarters/blob/30d766d289e256b9e2cd6e1f6876e4e266e8500b/fe-angular/init.sh#L21) folds the project name as an additional hierarchy level into the generated stuff. The purpose of this is to enable Angular to have multiple projects in one code base. OK, our quickstarter only generates one project, but what if someone wants to add more to the same code base? It would be good to not interfere with this possibility of maintaining stuff in parallel and let Angular Cli take care of everything.
If someone decides to run multiple projects in one Angular root, that person also needs to decide, how to deal with splitting or merging the coverage analysis on the different partial code bases.

PR #309, opened 2020-05-28
Add test coverage to Sonarqube for be-typescript-express

sino92:
https://github.com/opendevstack/ods-quickstarters/issues/213


sino92:
Screen Shot 2020-05-28 at 21 57 05
Tested with an already provisioned project:) My only regret is that I couldn't get it to test `quickstartWithJenkins` script. Ran into a snowball of issues there..

PR #312, opened 2020-05-31
Fix ionic unit test reporting #213

bljubisic:


bljubisic:
Ah, missed that one, will add it

fe-ionic/sonar-project.properties.template
```diff
@@ -7,8 +7,18 @@ sonar.projectName=@project_id@-@component_id@
...
 sonar.sourceEncoding=UTF-8
+
+# Exclude test files from coverage analysis
+sonar.coverage.exclusions=**/*.spec.ts
+
+# Exclude non type script files from coverage analysis
+sonar.exclusions=**/*.html,**/*.scss,**/*.json,**/*.ico,**/*.svg
+
+# Define coverage information inside the Angular project
+sonar.javascript.lcov.reportPaths=coverage/@component_id@/lcov.info
```
cschweikert, line 23: I guess ``@component_id/`` needs to be removed from the path. fe-ionic gets provisioned with the following ``karma.conf.js``: https://github.com/opendevstack/ods-quickstarters/blob/master/fe-ionic/files/karma.conf.js
This one points to the default setting for the ``coverageIstanbulReporter``, which usually is ``coverage/lcov.info``.
Please note: This is different from how it's done in fe-angular. See also https://github.com/opendevstack/ods-quickstarters/pull/306#issuecomment-634749519

fe-ionic/files/package.json
```diff
@@ -1,8 +1,6 @@
 {
-  "name": "ionic-quickstart",
-  "version": "0.0.1",
-  "author": "BIX",
-  "private": true,
```
ManuelFeller, line 5: Any reason you removed this ("private": true)?
Is no big deal, but this avoids an accidental publish to a package repository.
I am OK if we say this in the developers responsibility, just to make sure we want this (since we normally do not publish our full apps to a npm repo)...

bljubisic: I was just trying to be compliant to other fe repository, like fe-angular. Maybe we should keep this private setting

fe-ionic/sonar-project.properties.template
bljubisic: Thanks Christian 😉

fe-ionic/files/package.json
ManuelFeller: Then we just found something we can add to all other node based Quickstarters I guess? 😅

PR #313, opened 2020-06-01
Simplified example of spock test to run always

hrcornejo:
Removed geb tests to Google home page. Created dummies geb tests to avoid problems with proxies, changes, etc...


hrcornejo:
Fix: PLTF-1050

metmajer:
@martsec @hrcornejo The Google example was here for a purpose. It showed the use of a simple Page pattern. Why delete?

hrcornejo:
@metmajer because the page was giving errors:
- Processing JS
- Changing the page for another, the proxy was cutting the connection (not always but sometimes)
The sctructure of Geb test continues and the Google example only gives you a "Check title test". IMO not very important lesson to learn Geb.

PR #314, opened 2020-06-03
Build Jenkins slaves on GitHub Actions

michaelsauter:
* Moves all image build related files under `docker` in each slave directory (to avoid confusing stuff in `ocp-config` with image code)
* Drop `.rhel7` suffix from `Dockerfile`
* Sets the base image in each `Dockerfiles` to the prebuilt image to allow building using `docker build`
* Build images in Github Actions where it works out of the box (for the disabled NodeJS10 and Scala, we need to setup a Nexus instance)
* Use build args instead of env variables - this is potentially breaking, as noted in the update docs. Given we want to eventually move to pre-built images we cannot have cluster-specific variables in the images - users should not rely on those!
* Use `NEXUS_URL`/`context.nexusUrl` consistently throughout
* Simplify and align airflow setup script (@gerardcl what's the status on its removal?)

FYI @oalyman


common/jenkins-slaves/scala/sbtconfig/repositories
```diff
@@ -1,4 +0,0 @@
-[repositories]
-  local
-  my-ivy-proxy-releases: NEXUS_HOST/repository/ivy-releases, [organization]/[module]/(scala_[scalaVersion]/)(sbt_[sbtVersion]/)[revision]/[type]s/[artifact](-[classifier]).[ext]
-  my-maven-proxy-releases: NEXUS_HOST/repository/maven-public
```
michaelsauter, line 4: File wasn't actually deleted, just moved under `docker` (see above)

Issue #18, opened 2019-05-08
Angular / Node Builds are not distinguishing between DEV and PROD Environment


ManuelFeller:
Currently there are no environment dependent builds.
This prevents Angular Applications form being build with the `--prod` parameter, but might also affect other Quickstarters / Frameworks.

The first affected place I would suggest to make environment aware is
https://github.com/opendevstack/ods-project-quickstarters/blob/4446edf4a8929474035f781d91e99a771f20ac8e/boilerplates/fe-angular/Jenkinsfile#L38

There adding another type of build call e.g. `yarn buildprod` in case of a release environment could help.
That then needs to be added to the according `package.json` of a project.
So it might have to be added to all the Quickstarters but would enable ODS to have environment specific configurations in the frontend, e.g. different tracking id's for development and production.

It needs to be evaluated what configuration option makes sense (use it or not, what branch type / name is seen as production, ...)

michaelsauter:
@ManuelFeller Could you add the suggestion by @SHoen to the `Jenkinsfile.template`? I think that would be a good improvement for 3.x.

ManuelFeller:
@michaelsauter Will do, sure

ManuelFeller:
@SHoen since by default we only have `master` and `develop` (no `prod`) branches I only check for master in my PR. Also, for being consistent, I used `'master'.equals(context.gitBranch)` as it was already done in the Ionic Quickstarter.
Still thanks for reminding me with your suggestion that I opened this quite some time ago ;-)

PR #320, opened 2020-06-05
Master branch detection for production builds in Angular

ManuelFeller:
Adding production build detection from ionic quickstarter to the angular quickstarter, solving issue #18

### DIFF
--- a/fe-angular/Jenkinsfile.template
+++ b/fe-angular/Jenkinsfile.template
@@ -20,7 +20,11 @@ def stageBuild(def context) {
   stage('Build') {
     withEnv(["TAGVERSION=${context.tagversion}", "NEXUS_HOST=${context.nexusHost}"]) {
       sh "npm install"
-      sh "npm run build"
+      if ('master'.equals(context.gitBranch)) {
+        sh "npm run build --prod"
+      } else {
+        sh "npm run build"
+      }
     }
     sh "cp -r dist/${context.componentId} docker/dist"
   }


cschweikert:
@ManuelFeller Mh, I guess this change is not necessary: the prod flag is always attached from within the ``package.json``. See also https://github.com/opendevstack/ods-quickstarters/blob/master/fe-angular/files/package.json. Maybe it also failes now on master since it will attach ``--prod`` twice?

cschweikert:
In addition I would suggest to always build as production build. I had cases in the past, where certain unwanted behavior only occured due to the production build, e.g. caused by the uglification.

ManuelFeller:
I get the point @cschweikert, and I can remove that again (was somehow unaware of that change; I think when I opened that issue this was not integrated in the `package.json` - and I forgot to check there since I normally do not put the production flag in the `package.json` - a habit from Ionic projects).

Maybe you can add your points here a bit more detailed (performance judging, minification and E2E issues you experienced) and then @bljubisic, @netzartist, @rianet, @m-apsolon, @herrkoch, @akhilsoman and @ultrasonicsoft can let us know if they see it the same way or if they want to have dev builds for a reason.

cschweikert:
As @ManuelFeller mentioned there are pros and cons when deferring between dev and prod build.
Pros:
 * faster builds in dev mode
 * source maps/better readable code in the hosted application, that was build in dev mode
Cons:
 * final performance from code optimization can only be experienced in the prod version on master
 * dev and prod builds are not identical. This will increase the risk of issues that will only appear in a production build from master.

Example: I once did the mistake and relied on the name property of a class, which was turned into a random letter due to uglification 😅

I would suggest to start with production builds from all branches to not be surprised by any branch-specific build behavior. If someone needs special treatment for dev branches, e.g. source maps, ``angular.json`` offers the possibility to create multiple build configurations with many different things to configure (uglification, source maps, budgets, etc.). Setting up something branch-specific would then be the respective developer's responsibility.

An alternative to this could be to call the production build with the additional parameter ``--sourceMap=true``. Less invasive, but with the greatest benefit of the dev build from above.

What do you think? :)

ManuelFeller:
I would then go with prod builds and sourceMap=true.

This will also make sure that you run into resource issues with your build pod early (I had the case that dev worked but prod consumed quite a bit more and thus always broke because the pod ran out of memory).
In case we decide for this way I will recommendations / reasons from this thread to the `readme.md` of the generated project...

Issue #315, opened 2020-06-03
Switch from typescript to Js for node-express quickstarter


sino92:
**Is your feature request related to a problem? Please describe.**
Yeoman [generator](https://github.com/opitzconsulting/generator-node-express-typescript) isn't maintained frequently and makes it more difficult to maintain this quickstarter. 
Is anyone in favor of switching to another scaffolding tool, sticking to this one or quickstart with a clean JS express [generator](https://expressjs.com/en/starter/generator.html)? 

Here's a related [suggestion](https://github.com/opendevstack/ods-quickstarters/issues/88#issuecomment-582464267) from the past by @velcrin

michaelsauter:
I found https://www.npmjs.com/package/express-generator-typescript. Maybe also worth a try?

akhilsoman:
> I found https://www.npmjs.com/package/express-generator-typescript. Maybe also worth a try?

Yes, this looks promising. I have tested it locally and it generates a very minimal express application with typescript

sino92:
Nice, same as @akhilsoman suggested earlier. Would be curious to hear out what others think @ManuelFeller @m-apsolon @rianet @ultrasonicsoft @bljubisic @rattermeyer @velcrin @cschweikert @herrkoch @netzartist

bljubisic:
I have not tried it, but I will in next couple of days. I am all for typescript though :)

velcrin:
I also stumbled upon [express-generator-typescript](https://www.npmjs.com/package/express-generator-typescript) and it seems pretty opinionated. It does not only set up express and ts, it adds concepts such as `entities`, `logging`, `dao`, `multi env`, ... Do we want to go there a quickstarter?

netzartist:
Maybe this would fit, too (no experience with it): https://github.com/bitjson/typescript-starter

cschweikert:
* express-generator-typescript: looks nice except the dao and entities as @velcrin mentioned. Coverage analysis support might need to be added.
* typescript-start: doesn't preinstall anything express-related. Beside that it looks nice, presets a lot of usefull things like coverage analysis, prettier, vscode debug support, ...

Either we stick to a good but not perfect generator solution, or we provide and maintain a full example app ourself. I guess express-generator-typescript is a quick win for now.

velcrin:
[...]

PR #326, opened 2020-06-15
Remake node express typescript quickstarter with minimal dependencies

sino92:
README update can be found [here](https://github.com/opendevstack/ods-quickstarters/pull/327)
## Related issues
#315 
#95


be-typescript-express/files/README.md
```diff
@@ -0,0 +1,33 @@
+# be-typescript-express
+
+## Project setup
+
+### Configure .npmrc
```
cschweikert, line 5: I don't know, if a section about the ``.npmrc`` file is actually needed. The quickstarter should work without this file. If this file is needed, e.g. for using non-public libraries, it would recommend to add it via ``Jenkinsfile`` and Jenkins Credentials plugin like this:
```
def stageBuild(def context) {
  stage('Build') {
    withCredentials([file(credentialsId: 'npmrc', variable: 'NPMRC')]) {
      sh "cp $NPMRC .npmrc"
      sh "npm install"
      ...
```
This would help to keep the nexus credentials out of the codebase.

michaelsauter: @cschweikert How would the dependencies be installed on the local machine then? Shouldn't they be pulled via Nexus as well?

be-typescript-express/init.sh
```diff
@@ -1,32 +1,30 @@
...
 
-echo "copy fixes from quickstart to generated project"
-cp -r $SCRIPT_DIR/fix/. src/
+echo "Generate ts config file. Skip library check and redirect transpiled ts files to dist folder."
+npx tsc --init --skipLibCheck --outDir "./dist"
 
-echo "adding directories 'lib' and 'docker' to nyc exlude list to fix coverage test issue"
-sed -i 's/--reporter=text/--exclude=\\"docker\\" --exclude=\\"lib\\" --reporter=text/g' package.json
+echo "Add npm start to standard package json"
+python $SCRIPT_DIR/add-to-package-json.py $SCRIPT_DIR/files/package.json
```
michaelsauter, line 30: I wonder if it is worth having this file? Maybe it would be simpler to just execute the commands here from the `Jenkinsfile`?

be-typescript-express/add-to-package-json.py
```diff
@@ -0,0 +1,24 @@
...
+                "test": "JEST_JUNIT_OUTPUT_DIR='build/test-results/test' JEST_JUNIT_OUTPUT_NAME='test-results.xml' npx jest --reporters=default --reporters=jest-junit --coverage --coverageDirectory=coverage_output --forceExit ./dist"
+            }
+        )
+
+    with open(packageJson, 'w') as file:
+        file.write(json.dumps(obj, indent=4, sort_keys=True))
+        file.close()
+
+if __name__ == "__main__":
+    appendCommandsToPackageJson(sys.argv[1])
```
michaelsauter, line 24: I think this could be done in Groovy, right from the `Jenkinsfile`. Might be a bit easier as we only have Groovy code then, not Groovy and Python ... What do you think?

sino92: I essentially wanted to keep the Jenkinsfile short and hence we ended up with an `init.sh` and a python file. The code in both scripts is easy to understand but I get the upside of having the logic written in Groovy. If a longer Jenkinsfile is ok for you, then sure

be-typescript-express/init.sh
```diff
@@ -1,32 +1,30 @@
...
-npm install typescript@3.2.1 --save-dev
-npm install jest -g
-npm install jest-junit --save-dev
+npm i typescript@$TYPESCRIPT_VERSION express@$EXPRESS_VERSION @types/node@$NODE_TYPE_VERSION jest@$JEST_VERSION @types/jest@$JEST_VERSION
+npm i jest-junit@$JEST_JUNIT_VERSION --save-dev
 
-echo "copy fixes from quickstart to generated project"
-cp -r $SCRIPT_DIR/fix/. src/
+echo "Generate ts config file. Skip library check and redirect transpiled ts files to dist folder."
+npx tsc --init --skipLibCheck --outDir "./dist"
```
akhilsoman, line 27: I am not really sure but the i think the previous generator used output build files to a folder called '/lib'.  You might need to update the docker [file](https://github.com/BIX-Digital/ods-quickstarters/blob/fcc54cad3058e898bd4084d387465cf22e05a88a/be-typescript-express/files/docker/Dockerfile#L13)  to point to this new path to boot the application.

be-typescript-express/files/README.md
cschweikert: If you don't configure any registry (e.g. via ``.npmrc`` file), the dependencies are gathered from https://registry.npmjs.org directly. The dependencies inside the ``package.json`` of this quickstarter are available via https://registry.npmjs.org. Nexus usually works as a proxy to that registry.

Since we have a Nexus instance already in place, it would make sense to actually use it. AFAIK we are not setting or copying something (e.g. a ``.npmrc`` file) that would configure Nexus to be used automatically in all of our npm- or yarn-based builds.

If a developer wants to use this Nexus instance, configuring this via the ``.npmrc`` file locally is a good way to go. But a developer should NOT commit such a file, since it contains secrets. For this it would make sense to put it to the ``.gitignore``. Since it now only exists locally it also needs to be injected as a secret during a build process like mentioned above.

sino92: Hmm, then it would make sense then to rename the header to `Configure .npmrc locally` instead and the comments about npmrc configs so that it addresses a wider audience.
```
Configure your .npmrc file to download the npm dependencies from your private remote repository.
```
As for the Jenkinsfile, I'll update it as well...

michaelsauter: @ManuelFeller If I got your PR #287 right, then the agent image actually ships with a global npmrc pointing to Nexus, right?

@rianet If I remember correctly, you had a different setup in our last project, right? Didn't you use env vars in the committed `.npmrc`?

I think it would be great to have one unified handling of `.npmrc` across all the quickstarters ... as far as I understand it our choices are either a) commit `.npmrc` WITHOUT credentials or b) ignore `.npmrc` in Git but help the user to set one up locally. Right?

ManuelFeller: @michaelsauter correct, we now have the two node slaves with the matching `.npmrc` in the slave itself.

**But:** we either need to agree on all devs having this setting too on their dev machines (else the `package-lock.json` will point to the official registry **AND** need to be aware that this does not solve the issue for Ionic since they build on their slaves, so we need to keep it there.
So maybe we keep it for now to make sure nothing breaks when you are not aware of this (if you have the non-nexus url in the lockfile the slaves `.npmrc` will also not help - and that may lead to the proxy issues again).

_ToDo:_ documentation and alignment... ;-)

ManuelFeller: Another note on the question how @rianet did it in her last project:
yes, they used env-vars since this way also works with Ionics build services.

It could be a good idea to have some guidance how to do so locally and have it prepared in all containers that builds node stuff...

sino92: Then I think it makes sense to remove `.npmrc` config from README as we're not dependent on any internal libs like you said @cschweikert . @michaelsauter  If I understood you correctly, when you said you wanted a unified way of handling `.npmrc` across all quickstarters, you favor this one to also use `.npmrc`? Is this to make it future proof in case developers add libs from internal registry?

ManuelFeller: It's not that easy.

I we remove it here we need to have it documented (and in our company communicate) that all users developing node based projects with ODS **must** use a local, system wide `.npmrc` pointing to our nexus. (Or have one not checked in to the projects.

The image for building node apps (node 10 and node 12) have this active, using the configured internal nexus. But if the developer does not have the same system wide `.npmrc` enabled for his system (and also no local `.npmrc` in the project) npm will install from the public registry and write these URL's of the packages into the `package.lock.json`. And that file overrides any registry url set elsewhere -> builds will then not use the nexus repository.

Also, we do not control the Ionic build environment, so there for sure we need to keep the `.npmrc` in the codebase.

What we could do is having a section in the readme that explains how to configure your local machine to use the nexus that is configured in the used ODS installation and teach how important that is before you execute a `npm install` (thus generating the `package-lock.json`)

As said before, I would not do any changes to the way how we treat `.npmrc` in this with PR but come up with a more complete solution regarding all Quickstarters that use node based tech. Because for sure we need to communicate / document where we differ from the everyday node development a new developer is most likely used to...

be-typescript-express/files/src/index.ts
```diff
@@ -0,0 +1,7 @@
+var express = require("express")
+
+const app = express();
+
+app.listen(3000, () => {
```
akhilsoman, line 5: Hey @sino92 , it would be also better to add a simple response to the root url. It is the easiest way to test the quickstarter after it is deployed. I guess otherwise it will responds with a  'Cannot get /' message by default.

something like this will be enough i think,
`app.get('/', (req, res) => res.send('Hello World!'))`

be-typescript-express/files/README.md
cschweikert: Agreed ;)

Maybe we also add `.npmrc` into the `/be-typescript-express/files/.gitignore`. This will prevent any accidential commits of the `.npmrc` file.

be-typescript-express/add-to-package-json.py
sino92: @michaelsauter Can't read a file with method without script approval unfortunately. 
`Scripts not permitted to use new java.io.File java.lang.String`
Can we keep the python script? It's also easier for me to test and maintain it as is.

be-typescript-express/files/README.md
sino92: I am resolving this conversation. I think a separate issue on this would be highly beneficial to keep some quickstarters consistent across guidelines/documentation around `.npmrc`. We can always update this quickstarter once we reach a consensus there. 
Since libraries we install for the quickstarters are available in the official registry, changes regarding `.npmrc` shouldn't be a blocker against this PR.

be-typescript-express/add-to-package-json.py
michaelsauter: Sure we can keep it. It should be very easy to read the JSON in Groovy though: https://www.jenkins.io/doc/pipeline/steps/pipeline-utility-steps/#readjson-read-json-from-files-in-the-workspace. Then you simply add the key, and use `writeFile` to write. But this was just a suggestion, I don't mind either way.

be-typescript-express/files/README.md
rianet: Correct, for our Ionic project we commit ```.npmrc``` with env vars, we use the secrets in ionic dashboard to build and package the app using nexus, not the public npm registry. But this is a very specific use case for Ionic and the usage of Ionic Dashboard.

For other npm based projects ignoring the file and raising the awareness to set it up locally using nexus or your own registry it's fine. However, what I like about the approach using env vars is that as a developer when you clone your repo and ```npm install``` you are reminded to set up your local env vars and the usage of nexus.

be-typescript-express/add-to-package-json.py
sino92: Ah nice, didn't know that read/write JSON methods existed in Jenkins. It works now.

be-typescript-express/Jenkinsfile
```diff
@@ -1,23 +1,37 @@
...
+      def expressVersion = "4.17.1"
+      def jestVersion = "26.0.0"
+      def jestJunitVersion = "10.0.0"
+
+      sh "npm i typescript@$typescriptVersion express@$expressVersion @types/node@$nodeTypeVersion jest@$jestVersion @types/jest@$jestVersion"
+      sh "npm i jest-junit@$jestJunitVersion --save-dev"
+
+      print "Generate ts config file. Skip library check and redirect transpiled ts files to dist folder."
+      print "Target 10.0.0 and newer versions. Tell the compiler its okay to output JS syntax features from ES2018."
+      sh "npx tsc --init --lib 'es2018' --target 'es2018' --skipLibCheck --outDir './dist'"
```
cschweikert, line 30: veeeery cool :)

sino92: thanks for the help :)

PR #327, opened 2020-06-15
Update node express typescript quickstarter readme

sino92:


docs/modules/quickstarters/pages/be-typescript-express.adoc
```diff
@@ -22,35 +22,42 @@ The files are generated using a https://yeoman.io/[yeoman] generator for https:/
...
+. http://www.typescriptlang.org/[Typescript ({caret}3.9.5)]
+
+== Libraries
+On top of the mentioned frameworks above, we use libraries:
+
+. https://www.npmjs.com/package/jest-junit/[Jest Junit ({caret}10.0.0)]
+. https://www.npmjs.com/package/@types/node[@types/node ({caret}14.0.13)]
+
+== Node version
+As our builder slave uses node v10, we strongly recommend you to use `v10.15.3`.
```
cschweikert, line 46: ``Dockerfile`` also uses Node v10 directly (``node:10-alpine``). I would rename it, e.g. like this:
```
As our builder slave and the `Dockerfile` uses node v10, we strongly recommend you to keep this in sync with the version you are developing for.
```
In principle using language specifics from newer node versions works as long as the ``Dockerfile`` also points to the newer node version. The only thing that might crash are unit tests. Those are running within the node v10-based slave during a Jenkins build.

sino92: Then to add from your previous comments, how about rephrasing it to: 
```
For local development, keep in mind that Dockerfile not only compiles your whole source code with node v10. 
As for when you push your code, your test code will be compiled using using node v10. This might cause some problems if you write and checked your tests using a different node version locally.
Let's say you push your latest test codes validated locally with node v12. In this case, Jenkins might complain about some compilation errors while its running those tests. To avoid such situations, we strongly recommend you to double check your test code to be in sync with node v10 before you push it.
```

cschweikert: The ``Dockerfile`` is not used for compiling. It is used for running the express server. How about this:
```
When you push your code, your runtime and test code will be compiled using node v10. After deployment the compiled code also runs in node v10. This might cause some problems, if you developed your code and tests using a different node version locally.

We strongly recommend you to double check your typescript compiler options in ``tsconfig.ts`` to be compatible with node v10, e.g. with a setting like this:
{
  "compilerOptions": {
    "lib": ["es2018"],
    "module": "commonjs",
    "target": "es2018"
  }
}
```

@sino92 How does the ``tsconfig.ts`` generated from the quickstarter look like?

sino92: Ah right, just used to run the express server, I assumed the build step was also there...
I like it. Let's keep the first paragraph. How about we add those in `tsconfig` file and just mention why the setting is made that way?
```
This is why we made the `tsconfig.ts` to be compatible with node v10. 
So bear in mind any changes to these compiler options might result in 
compilation and runtime errors.
```
So far we use `tsc --init --skipLibCheck --outDir './dist'` to generate tsconfig. A sample is can be found [here](https://bitbucket.bix-digital.com/projects/SINANPLAY/repos/sinanplay-customized-node-4/browse/tsconfig.json).

cschweikert: I like it. BTW: I did it wrong. It is `tsconfig.json` not `tsconfig.ts`.

The `tsconfig.json` you attached looks pretty much backwards compatible. It might even run on older node versions.

sino92: Here's an [update](https://github.com/opendevstack/ods-quickstarters/pull/326/commits/a6f9178198e43673c21dd171533017a9e89bcf6a) to tsconfig init.

PR #332, opened 2020-06-17
Feature/fail r shiny if packages not installed

### LINKED ISSUE #331: fail R-Shiny build if app.R dependencies are not found/installed
When building R-Shiny quickstarter's Dockerfile, the build does not fail if a package does not exist, which in turn can lead in a non-loading app

Fail the build/stage if any depencendy is missing

Keep as is right now and do not fail when dependencies are missing

**Additional context**
User will be always to define the behaviour of the build later on.
Relates to #330

gerardcl:
fixes #331
followup of #330


michaelsauter:
I don't know the tools, but it looks good that it should fail now :)

PR #358, opened 2020-07-07
Use new image import strategy if possible

michaelsauter:
See https://github.com/opendevstack/ods-jenkins-shared-library/pull/400.

@gerardcl Maybe you can check quickstarters you own if it would be possible to add this as well? So far I just did it for the obvious / easy ones :)

### DIFF
--- a/be-gateway-nginx/Jenkinsfile.template
+++ b/be-gateway-nginx/Jenkinsfile.template
@@ -10,7 +10,9 @@ odsComponentPipeline(
   ]
 ) { context ->
 
-  odsComponentStageBuildOpenShiftImage(context)
+  odsComponentStageImportOpenShiftImageOrElse(context) {
+    odsComponentStageBuildOpenShiftImage(context)
+  }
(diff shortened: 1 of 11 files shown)


gerardcl:
by the way, I tested successfully `ds-jupyter-notebook`, `ds-rshiny` and `be-gateway-nginx`

michaelsauter:
@gerardcl Hmm, I would think that `resourceName` is actually supported, see https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/vars/odsComponentStageImportOpenShiftImageOrElse.groovy#L30, https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/vars/odsComponentStageImportOpenShiftImageOrElse.groovy#L54 and https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/vars/odsComponentStageImportOpenShiftImageOrElse.groovy#L69. Did you attempt and it didn't work?

PR #365, opened 2020-07-09
Add golang smoke tests

georgfedermann:


tests/be-golang-plain/jenkinsfile_test.go
```diff
@@ -0,0 +1,129 @@
...
+		ImageTags:         []coreUtils.ImageTag{{Name: componentId, Tag: "latest"}},
+		BuildConfigs:      []string{componentId},
+		DeploymentConfigs: []string{componentId},
+		Services:          []string{componentId},
+		ImageStreams:      []string{componentId},
+	}
+
+	coreUtils.CheckResources(resourcesInTest, t)
+
+}
```
michaelsauter, line 129: @clemensutschig Wondering - did you decide on purpose against a table test that just runs the same code against different directories / golden files? The source code here looks largely identical to the other quickstarters ...

clemensutschig: @michaelsauter i thought about this as well.. but as quickstarters grow and Change i fehlt the couple lines of code dont matter..

michaelsauter: I see. However, right now I fail to see how we actually would need different code:

* Provisioning is a standardised process - calling a Jenkins Pipeline with some params. Right now the params are fixed, but even if they are more flexible, it would be easy to cover that in a table driven test
* Building a component is a standardised process - calling a Jenkins Pipeline with no params
* As output, there will always be stages as long as we use Jenkins, and SonarQube (if applicable to the QS) also returns the same thing for every QS (as every QS is equal to a SQ project)

The only thing were quickstarters could differ:
* Created resources
* Created files in the resulting repository

Those would be easy to cover in table driven tests. If not enough, we could even pass functions that do special things in the table driven tests to retain complete flexibility ...

I think we should think about it - as, at least right now, a "QS test" would only need to be:
* a name
* golden files

Would be easier to create / maintain?

PR #368, opened 2020-07-10
Clean up image tags

michaelsauter:
FYI @clemensutschig @georgfedermann 

This should ensure that you get the same Jenkins stages when you run the same Git commit SHA again (otherwise you see that the built image is re-used).

### DIFF
--- a/tests/utils/command-utils.go
+++ b/tests/utils/command-utils.go
@@ -1,38 +1,26 @@
 package utils
 
 import (
-    "bytes"
-    "fmt"
-    "os"
-    "os/exec"
-    "path"
-    "runtime"
+	"bytes"
+	"fmt"
+	"os"
+	"os/exec"
+	"path"
+	"runtime"
 )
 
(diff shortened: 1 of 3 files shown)


tests/utils/command-utils.go
```diff
@@ -1,38 +1,26 @@
...
-    cmd.Stderr = &stderr
-    err := cmd.Run()
-    return string(stdout.Bytes()), string(stderr.Bytes()), err
+	cmd := exec.Command(command, args...)
+	cmd.Env = append(os.Environ(), envVars...)
+	var stdout, stderr bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+	err := cmd.Run()
+	return string(stdout.Bytes()), string(stderr.Bytes()), err
```
michaelsauter, line 25: Those changes are only due to `go fmt` running automatically on my machine - make sure to run that :)

Issue #380, opened 2020-07-15
Update Ionic Quickstarter to @ionic/cli


ManuelFeller:
**Is your feature request related to a problem? Please describe.**
The old `@ionic` package is deprecated; `@ionic/cli` is the successor and should be used.
Also more consistent use of yarn / npm would be good...

**Describe the solution you'd like**
Update to the latest CLI that works in the build image and fully switch to npm

**Describe alternatives you've considered**
staying on an outdated version..?

**Additional context**
-

ManuelFeller:
So while working on this and testing with `@ionic/cli` in version 6.3.0 on node version 10.15.1 two things that I was not aware of came up:

- the used Angular version does no longer obey the `--single-run` parameter on `ng test`, leading to a not closing chrome -> does not work with CI out of the box
- I did not find a way (yet) to make a non-interactive install with a cordova integration since even the `ionic start yourproject blank --cordova` leads to the following interactive request:

```
[WARN] About to integrate your app with Cordova.
       
       We now recommend ⚡️ Capacitor⚡️  (https://ion.link/capacitor) as the official native runtime for Ionic. To
       learn about the differences between Capacitor and Cordova, see these docs[1]. For a getting started guide, see
       these docs[2].
       
       [1]: https://ion.link/capacitor-differences-with-cordova-docs
       [2]: https://ion.link/capacitor-using-with-ionic-docs
       
? Are you sure you want to continue? (Y/n)
```
When you try `--no-interactive` added to the call this will be answered with no, leading to a not executed integration.

I see two possible solutions for new Ionic apps:
Not adding any integration and maybe adding the instructions how to start the integration (run `ionic integrations enable cordova` / `ionic integrations enable capacitor`) to the `README` - or going with capacitor (which may lead to issues later since not all Cordova plugins work with it for sure).

I personally vote for keeping the integration open solution since it allows the project to choose - what do you think @cschweikert @rianet @bljubisic 

And regarding the karma not closing Chrome issue I will check what solutions could work, maybe I go for a customized karma config or something like that...

cschweikert:
Have you tried running it with ``CI=true npm run ...`` or so on your local machine? The environment variable ``CI`` is usually used to determine CI/CD tools running the script, which again leads to a different behavior. ``ionic srtart`` for example doesn't ask any questions anymore, if you apply ``CI=true``.

cschweikert:
[...]

PR #382, opened 2020-07-15
fe-ionic smoke tests

cschweikert:
smoke tests for quickstarter fe-ionic

**Please note: The provisioning of fe-ionic is currently failing. Related issues are #381 and #380. When we merge this PR, the smoke tests will also fail as long the issues above were not fixed.**


clemensutschig:
@cschweikert  - so this one works then?

cschweikert:
I need to test it

cschweikert:
#381 and #380 are sorted out. There was a follow-up fix #390 that is also merged now. I tested the fe-ionic smoke test and it finally has passed :)

PR #385, opened 2020-07-15
WIP: add jupyter smoketests

georgfedermann:


ds-jupyter-notebook/Jenkinsfile
```diff
@@ -16,7 +16,7 @@ odsQuickstarterPipeline(
       --component ${context.componentId} \
       --nexus-url ${context.nexusUrl} \
       --nexus-username ${context.nexusUsername} \
-      --nexus-password ${context.nexusPassword} \
+      --nexus-password b3BlbnNoaWZ0 \
```
georgfedermann, line 19: this hack still requires fixing ...

clemensutschig: @georgfedermann  - what's the problem here? the password must be set - I guess it#s not base64 encoded, right?

PR #386, opened 2020-07-16
380 ionic cli package

ManuelFeller:
Pull request for Issue #380 
- updated Ionic CLI package
- removed yarn usage to use npm consistently 
- removed `package.json` from `files` dir to use the generated default one
- removed `karma.conf.js` from `files` dir to use the generated one
- added replacement scripting for `karma.conf.js` in `init.sh` similar to the Angular Quickstarter


fe-ionic/init.sh
```diff
@@ -18,11 +18,56 @@ while [[ "$#" > 0 ]]; do case $1 in
 esac; shift; done
 
 echo "generate project"
+<<<<<<< HEAD
```
clemensutschig, line 21: @ManuelFeller  - merge conflict? ....

ManuelFeller: checking right now, had solved that already locally (I thought)

ManuelFeller: OK, no longer in there now, added some explanations about what is happening

fe-ionic/init.sh
```diff
@@ -18,11 +18,58 @@ while [[ "$#" > 0 ]]; do case $1 in
 esac; shift; done
 
 echo "generate project"
+# create temp-dir since non-interactive ionic-cli can not generate into existing directories
 mkdir start_$COMPONENT
 cd start_$COMPONENT
-ionic start $COMPONENT blank --type=angular --no-deps --no-git
+
+# creating project
+ionic start $COMPONENT blank --type=angular --no-deps --no-git --no-interactive
```
cschweikert, line 26: ``--no-interactive`` is not needed. Since ``CI=true`` it has the same behavior.

fe-ionic/init.sh
```diff
@@ -18,11 +18,58 @@ while [[ "$#" > 0 ]]; do case $1 in
...
+
+# move generated project to the intended directory
 mv $COMPONENT/ ..
+# switch to component directory
 cd ../$COMPONENT
+# remove empty temp-dir
+rm ../start_$COMPONENT
+
+echo "Change test setup to single run in karma.conf.js"
```
cschweikert, line 35: ``fe-ionic/files/package.json`` was removed. On the bottom of this ``init.sh`` is still a ``rm ./package.json``. After this there wouldn't be any package.json left...

beside that the ``fe-ionic/files/package.json`` had some finetuning on the tasks like added parameters for generating code coverage analysis data:
```
"test": "ng test --code-coverage --reporters=junit",
```

fe-ionic/init.sh
ManuelFeller: Idea behind is to have same behaviour reproducible in your local system without having anything to remember. But can remove if wanted...

fe-ionic/init.sh
ManuelFeller: Done as requested

PR #397, opened 2020-07-20
jenkins agent python pip using nexus as default index

### LINKED ISSUE #396: set nexus as default pip repo index for jenkins python agent
When using the jenkins agent python based one always needs to remember to define which index to reference in order to assure using nexus.

To have pip already fully and properly configured (e.g.: having as an extra index the official pypi repo) within the jenkins python agent.

Keep current implementation and assure/remember that everyone (quickstarters and users) keep using nexus.

**Additional context**
Would be nice to also update the related Jenkinsfile.templates of the python quickstarters.

gerardcl:
fixes #396

### DIFF
--- a/be-python-flask/Jenkinsfile.template
+++ b/be-python-flask/Jenkinsfile.template
@@ -31,7 +31,7 @@ def stageTestSuite(def context) {
     sh """
       virtualenv testsuite
       . ./testsuite/bin/activate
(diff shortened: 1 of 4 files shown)


gerardcl:
it seems that the tests are not using/having the ods-core.env params required for building the jenkins python agent...will take a look closer to it!

gerardcl:
ok as discussed with @clemensutschig:  I will make nexus optional build arg for jenkins python agent

clemensutschig:
@gerardcl  - correct, only the ami tests - bootstrap nexus - so the default test has to be without (see also the failing check)

PR #408, opened 2020-07-24
use python rh based packages and keep rh development packages

### LINKED ISSUE #407: python agent should build lib packages for python packages that need compiling
**Describe the bug**
When installing python packages that need to compile C/C++ code the current jenkins agent fails.

**To Reproduce**
1. Provision a be-python-flask
2. Add a python package in `requirements.txt` like `pyspark`
3. Commit and push
4. Pipeline fails

**Expected behavior**
Python agent should be able to compile source code when installing python packages

 - OpenShift: all
 - OpenDevStack: 2.x, master

```
 gcc -pthread -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -std=c99 -fPIC -I/tmp/workspace/projectx-cd/projectx-cd-feature-test/testsuite/include -I/usr/local/include/python3.6m -c ./snappy/snappy.cc -o build/temp.linux-x86_64-3.6/./snappy/snappy.o -I./leveldb/include -I./leveldb -I./snappy -I. -fno-builtin-memcmp -O2 -fPIC -DNDEBUG -DSNAPPY -pthread -Wall -DOS_LINUX -DLEVELDB_PLATFORM_POSIX
  unable to execute 'gcc': No such file or directory
  error: command 'gcc' failed with exit status 1
```

**Additional context**
This was never happening since most of the times users use conda based environments.

gerardcl:
fixes #407
- we do not need to compile anymore python since we use RetHat based packages 👍 
- this also fixes possible airflow requirements when migrating to 3.x (users that might still use airflow they now can always keep upgrading ODS by using the python agent instead 🚀 )
- fix for 2.x will not be provided since there has not been any requirement as of today and users could always use the airflow based one

### DIFF
--- a/common/jenkins-agents/python/docker/Dockerfile
+++ b/common/jenkins-agents/python/docker/Dockerfile
@@ -5,45 +5,42 @@ LABEL maintainer="Gerard Castillo <gerard.castillo@boehringer-ingelheim.com>"
 ARG nexusHost
 ARG nexusAuth
 
-ENV PYTHON_VERSION=3.6.10
-ENV VENV_VERSION=20.0.9
-ENV INSTALL_PKGS="yum-utils gcc make openssl-devel zlib-devel sqlite-devel"
+ENV LANGUAGE en_US.UTF-8
+ENV LANG en_US.UTF-8
+ENV LC_ALL en_US.UTF-8
+ENV LC_CTYPE en_US.UTF-8
+ENV LC_MESSAGES en_US.UTF-8
+ENV PATH=/opt/rh/rh-python36/root/usr/bin${PATH:+:${PATH}} \
+    LD_LIBRARY_PATH=/opt/rh/rh-python36/root/usr/lib64${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}} \
+    MANPATH=/opt/rh/rh-python36/root/usr/share/man:$MANPATH \
+    PKG_CONFIG_PATH=/opt/rh/rh-python36/root/usr/lib64/pkgconfig${PKG_CONFIG_PATH:+:${PKG_CONFIG_PATH}} \
+    XDG_DATA_DIRS="/opt/rh/rh-python36/root/usr/share:${XDG_DATA_DIRS:-/usr/local/share:/usr/share}"
 
-RUN set -x \
-    && yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS \
-    && yum clean all \
-    && rm -rf /var/cache/yum/*
+# Enable rhel-server-rhscl-7-rpms repo
+RUN set -x && \
+    yum-config-manager --enable rhel-server-rhscl-7-rpms && \
+    yum-config-manager --disable rhel-7-server-htb-rpms && \
+    yum makecache
 
-RUN cd /tmp \
-    && curl -O https://www.python.org/ftp/python/${PYTHON_VERSION}/Python-${PYTHON_VERSION}.tgz \
-    && tar xzf Python-${PYTHON_VERSION}.tgz -C / \
-    && rm -rf Python-${Python_VERSION}.tgz
-
-RUN cd /Python-${PYTHON_VERSION} \
-    && ./configure \
-    && make altinstall \
-    && ln -s /Python-${PYTHON_VERSION}/python /usr/local/sbin/python3 \
-    && python3 -V \
-    && chmod a+rx /Python-${PYTHON_VERSION} \
-    && chmod a+rx /Python-${PYTHON_VERSION}/python \
-    && yum remove -y $INSTALL_PKGS
-
-RUN curl "https://bootstrap.pypa.io/get-pip.py" -o "get-pip.py" \
-    && python3 get-pip.py
+# Python 3.6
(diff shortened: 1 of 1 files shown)


common/jenkins-agents/python/docker/Dockerfile
```diff
@@ -5,44 +5,43 @@ LABEL maintainer="Gerard Castillo <gerard.castillo@boehringer-ingelheim.com>"
...
+    ENV=/usr/local/bin/scl_enable \
+    PROMPT_COMMAND=". /usr/local/bin/scl_enable"
+
+COPY contrib/bin/scl_enable /usr/local/bin/scl_enable
+
+# Enable rhel-server-rhscl-7-rpms repo
+RUN set -x && \
+    yum-config-manager --enable rhel-server-rhscl-7-rpms && \
+    yum-config-manager --disable rhel-7-server-htb-rpms && \
+    yum makecache
```
michaelsauter, line 28: Will this not fail on CENTOS?

gerardcl: I tested builds within openshift and locally building from the main jenkins agent base and it works :) it was already working before within the airflow agent in ODS2.x.
What I want to improve anyway is the usage of the scl_enable script, I think it might not be used, but I just got inspired from the nodejs agent one :)

michaelsauter: Ah nice. Talk to @oalyman regarding this `scl_enable` thingy, I remember he had questions about it when updating the Scala agent, and might have removed it completely.

henrjk:
Hi @gerardcl can you elaborate on why the switch to the rh package?

gerardcl:
hi @henrjk  sure! first it was something I already had in mind since many time ago but had not good chance to provide it. Second I believe it is better using already created and provided packages from RH than compiling ourself the whole python. I find it more stable and secure. Do you see any issue though?

gerardcl:
@henrjk  another question I just thought, shall we upgrade to python 3.8? should be stright forward and will bring nice things! :)

henrjk:
@geradcl the main potential issue could be that redhat would not update the packages frequently enough on rhel7. 

rhel7 is now under Maintenance support 1 and transition to Maintenance support 2 in August. (see https://access.redhat.com/support/policy/updates/errata/?extIdCarryOver=true&sc_cid=701f2000001Css5AAC#Life_Cycle_Dates). However I am not sure what the impact on the python packages will be.

common/jenkins-agents/python/docker/Dockerfile
```diff
@@ -5,45 +5,42 @@ LABEL maintainer="Gerard Castillo <gerard.castillo@boehringer-ingelheim.com>"
...
-    && rm -rf Python-${Python_VERSION}.tgz
-
-RUN cd /Python-${PYTHON_VERSION} \
-    && ./configure \
-    && make altinstall \
-    && ln -s /Python-${PYTHON_VERSION}/python /usr/local/sbin/python3 \
-    && python3 -V \
-    && chmod a+rx /Python-${PYTHON_VERSION} \
-    && chmod a+rx /Python-${PYTHON_VERSION}/python \
-    && yum remove -y $INSTALL_PKGS
```
henrjk, line 29: The `yum remove` is now missing. Will this make the image larger.

common/jenkins-agents/python/docker/Dockerfile
```diff
@@ -5,45 +5,42 @@ LABEL maintainer="Gerard Castillo <gerard.castillo@boehringer-ingelheim.com>"
 ARG nexusHost
 ARG nexusAuth
 
-ENV PYTHON_VERSION=3.6.10
```
henrjk, line 8: I liked having these env variables.

common/jenkins-agents/python/docker/Dockerfile
gerardcl: right, but that is by purpose because we want to address #407 and let users be able to build packages within jenkins job if they need/want to

common/jenkins-agents/python/docker/Dockerfile
gerardcl: yep, that was removed in favour of handling the versions by the package offered from redhat/centos

I agree that this current proposal might lead to less flexibility so I will continue working on this and test a new python compiled approach

hugowschneider:
Hi Guys, I think since we are running under OC, we should trust what RH provides.

On the one hand, RH has python 3.8 already available for usage; on the other hand, I would add the burden of maintaining our python compilation. We could also get an official python image if RH is the problem to avoid keeping our compilation.

See an official python image: https://github.com/docker-library/python/blob/44ded7c92b0c20abbeaa88785751eb0822d05ced/3.8/alpine3.12/Dockerfile

The Dockerfile is enormous, with lots of details that need understanding and need to be maintained. I wouldn't take this direction if it were my choice.

hugowschneider:
Some inspiration:

https://catalog.redhat.com/software/containers/rhscl/python-38-rhel7/5e838896dd19c7789613d625?container-tabs=dockerfile

gerardcl:
closing in favour of https://github.com/opendevstack/ods-quickstarters/pull/416

PR #416, opened 2020-07-28
Feature/python38 and deps upgrade for jenkins agent and quickstarters

### LINKED ISSUE #407: python agent should build lib packages for python packages that need compiling

### LINKED ISSUE #415: Lets upgrade to the latest python 3.8
The be-python-flask, ds-ml-service, ds-jupyter-notebook and the python agent are still on 3.6 and we should go to the latest 3.8 (currently 3.8.5).

I think for the deployed images these should be based on the [3.8.5-buster](https://github.com/docker-library/python/blob/44ded7c92b0c20abbeaa88785751eb0822d05ced/3.8/buster/Dockerfile) which should also allow for packages needing C compilation.  

The slim image could be used for having a more lightweight image, but I think allowing C packages is more important to most users

**Additional context**
Ideally this should go into ODS3

gerardcl:
fixes #407
fixes #415 

- all python quickstarters upgraded to python3.8 and also dependencies upgraded to latest versions  🚀
- jenkins python agent upgraded to python 3.8  🚀
- we do not need to compile anymore python since we use RetHat based packages 👍
- this also fixes possible airflow requirements when migrating to 3.x (users that might still use airflow they now can always keep upgrading ODS by using the python agent instead - tested with already existing use cases 😄 )

Tasks: 
- [x] Adjusted and verified tests in `tests/...` directory


michaelsauter:
@gerardcl Nice. This looks SOO much better.

I still have questions regarding the backwards compatibility though. If I have an existing Python component, then my `Dockerfile` will use e.g. Python 3.6, but once I'm pointing to the new Jenkins agent, that uses Python 3.8. Will that cause issues? Should it at least be mentioned in the update guides?

hugowschneider:
@michaelsauter Long story short, python guarantees this backward compatibility. The only thing I am not entirely sure if it is valid for libraries with C bindings, such as NumPy. But I am confident it should work in 99% of the cases when not in all. If it does not work, the libraries must be updated.

gerardcl:
thanks!! 👍 I tested that :)

Here notes for upgrade guide:
- ds-jupyter-notebook: not affected, they could keep using ODS2.x skeleton or just upgrade to 3.x. No issues except if it has been tweaked a lot, but that is obvious.
- ds-ml-service: there is a minor change required in the Jenkinsfile; basically installation of python packages within a jenkins job should be wrapped within a virtualenv (one can get inspired with the new Jenkinsfile.template)
- be-python-flask: there was already a breaking change a while ago -> https://github.com/opendevstack/ods-quickstarters/pull/152
- It is worth to mention too that there is no longer need to use nexus params within the jenkins jobs when using pip since it is already configured if it exists: https://github.com/opendevstack/ods-quickstarters/pull/397

NOTE: a change like this will always mainly affect projects that might be still using old dependencies -> which is forcing them to update for obvious reasons 🚀

michaelsauter:
@hugowschneider @gerardcl  Thanks!

Can you document this? The best way would be to create a partial like https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/docs/modules/jenkins-shared-library/partials/update-to-3x.adoc, and then we can reference this from the update guide page (in ods-core).

gerardcl:
will do later today 👍

be-python-flask/files/docker/Dockerfile
```diff
@@ -1,71 +1,24 @@
-FROM python:3.6.10-slim
+FROM python:3.8-buster
```
henrjk, line 1: Should this also be switched to use the rh-python packages so it is matching the agent?

gerardcl: In this case I would prefer (also since other users using the quickstarter noted this) the official python provided ones. On the ds python related ones we are already providing the centos7/rhel7 though. 
At the end one can switch the FROM to whatever preference/requirement.
On the other side, I do not see any issue on providing different FROMs in different places if at least all of them provide the python 3.8 runtime and dependencies.
So, in this case I would not switch.

gerardcl:
@henrjk  I just passed all tests related to the affected quickstarters on this PR with succes (thanks @cschweikert !!)
missing thing is to document the upgrade guide accordingly, will do later today or tomorrow early!

gerardcl:
@michaelsauter docs updated! I will merge during tomorrow morning if no issues!

docs/modules/jenkins-agents/partials/update-to-3x.adoc
```diff
@@ -1,3 +1,5 @@
...
+
+All python related quickstarters (`be-python-flask`, `ds-jupyter-notebook` and `ds-ml-service`) and jenkins agent `jenkins-agent-python` has been upgraded to python 3.8. This is not affecting default provisioned `Jenkinsfile` for `ds-jupyter-notebook` nor `be-python-flask` quickstarters that exist before ODS 3.x. But, `ds-ml-service` provisoned quickstarters before ODS 3.x need to slightly modify its `Jenkinsfile` before upgrading to ODS 3.x by just avoiding running any `pip install <..dependencies..> --user` (see new `Jenkinsfile.template` to get a https://github.com/opendevstack/ods-quickstarters/blob/master/ds-ml-service/Jenkinsfile.template[suggested fix], basically using `virutalenv`; which is the recommended way of working when requiring building python packages). Finally, if you provisioned `airflow-cluster` quickstarter, please note that it has been moved out from the officially supported quickstarters (`ods-quickstarters`) repository and moved to `extra-quickstarters`. Therefore, the `jenkins-slave-airflow:2.x` can be shifted to `jenkins-agent-python:3.x` when upgrading to ODS 3.x.
```
michaelsauter, line 5: Some minor remarks:

* Let's make this a separate section with its own header
* Let's make the part about airflow a separate section in itself
* There are a few typos e.g. `s/has been upgraded/have been upgraded`

PR #417, opened 2020-07-29
Fix failing test for git branches

georgfedermann:


georgfedermann:
@clemensutschig am on it

PR #428, opened 2020-08-03
Feature/cypress sonar testing

oalyman:
add sonar cube testing for cypress

### DIFF
--- a/e2e-cypress/Jenkinsfile.template
+++ b/e2e-cypress/Jenkinsfile.template
@@ -10,6 +10,7 @@ odsComponentPipeline(
   ]
 ) { context ->
   stageTest(context)
+  odsComponentStageScanWithSonar(context)
 }
 
 def stageTest(def context) {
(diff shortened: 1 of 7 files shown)


e2e-cypress/Jenkinsfile.template
```diff
@@ -10,6 +10,7 @@ odsComponentPipeline(
   ]
 ) { context ->
   stageTest(context)
+  odsComponentStageScanWithSonar(context)
```
michaelsauter, line 13: Should we reverse the order? I think it is probably not needed to run the tests before doing the analysis, right? And tests will likely take quite a bit longer ...

ManuelFeller: Agree with the reversed order. Running them after the SonarQube scan would also encourage you to do e2e tests locally to be more sure that that stage does not fail...

clemensutschig: does not work - for SQ you need binaries :(

michaelsauter: @clemensutschig This is TypeScript ... no binaries. IIRC SQ scanning works anytime against the source files only.

clemensutschig:
Not sure.. i had isses in the spock one.. but feel free to try.

Michael Sauter <notifications@github.com> schrieb am Mo., 3. Aug. 2020,
12:23:

PR #430, opened 2020-08-04
Bug/add ods namespace var

renedupont:
When I tried to run the quickstarter tests for our ods installation (which has not the default ods namespace name), it couldn't pull images in the jenkins job because it tried to pull it from <registry_ip>/ods/image:lastest.

I guess it is the same issue as we discovered yesterday in ods-core https://github.com/opendevstack/ods-core/commit/d9a7618c6ac576728dfbd199b56d8e7f3a9d9353#diff-007c127e12abf8044e814b788a553deaR71

Thus I added the variable to the tests.


renedupont:
added commit:

removes quality key from sonar check and adds count=1 to 'go test' to prevent caching

PR #447, opened 2020-08-21
Add health check for Go quickstarter

### LINKED ISSUE #137: Add health endpoint to Go quickstarter
Healthchecks should be configured out of the box, therefore the Go quickstarter should have a health endpoint.

Related to https://github.com/opendevstack/ods-quickstarters/issues/11.

michaelsauter:
Closes #137.

`common/ocp-config/component` is basically a copy of `common/ocp-config/component-environment`, with two changes:

* removed `env: ${ENV}` label from the pod template (it's not needed)
* added readiness probe (partially parameterised but with defaults)

The idea would be that every quickstarter migrates to this template now so that eventually we can remove `common/ocp-config/component-environment`.

### DIFF
--- a/common/ocp-config/component/template.yml
+++ b/common/ocp-config/component/template.yml
@@ -1,0 +1,195 @@
+apiVersion: v1
+kind: Template
+labels:
+  template: component-template
+  app: "${PROJECT}-${COMPONENT}"
+parameters:
+- name: PROJECT
+  displayName: Application
+  description: The name of the application project.
+  required: true
+- name: COMPONENT
+  displayName: Component
+  description: The name of the application component.
+  required: true
+- name: ENV
+  displayName: Environment
+  description: The environment stage.
+  required: true
+- name: MEMORY_LIMIT
+  displayName: Memory Limit
+  description: Maximum amount of memory available for the container.
+  value: 256Mi
+- name: MEMORY_REQUEST
+  displayName: Memory Request
+  description: Minimum amount of memory requested for the container.
+  value: 128Mi
+- name: CPU_LIMIT
+  displayName: CPU Limit
+  description: Maximum CPU (milli) cores available for the container.
+  value: 100m
+- name: CPU_REQUEST
+  displayName: CPU Request
+  description: Minimum CPU (milli) cores requested for the container.
+  value: 50m
+- name: TAGVERSION
+  displayName: tag version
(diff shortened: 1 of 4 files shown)


gerardcl:
that looks good! why not also adding liveness probe? any reason?

michaelsauter:
@gerardcl Hmm. My reasons not to add it:

* The concept of "liveness" is quite specific for an application, I think a quickstarter can probably not define what that means (might be different for e.g. Jupyter etc.)
* The main focus is to have more robust tests, which is solved by readiness probes alone

But open to be convinced otherwise :)

gerardcl:
let me try to convince you then hahah
- the reason I mention this is also that I see providing such params/templates/configs as also doing kind of pedagogic effort on pushing for good practices to any person using ODS
- since quickstarters are given as a default and simple skeleton there is no issue on providing the liveness probes IMHO, so they could even have the same config as the readiness in this case (then teams can anyway edit as per requirements of each deployment..)
- and, mixing both previous statements, we need anyway to document this in a way that developers will really understand why it is good/necessary, and even what each of both probes really mean and behave (relation with the deployment strategy timeout param, its defined timeouts, etc)

michaelsauter:
@gerardcl @henrjk OK you convinced me, let's add liveness probes from the start. Added in 9ce1de5.

common/ocp-config/component/template.yml
```diff
@@ -0,0 +1,195 @@
...
+- name: TAGVERSION
+  displayName: tag version
+  description: The version to be used.
+  value: latest
+  required: true
+- name: COMPLETION_DEADLINE_SECONDS
+  displayName: Completion Deadline Seconds
+  description: how long the docker build of the component can last before it is canceled
+  value: "1800"
+- name: READNINESS_PATH
```
henrjk, line 44: There is a type in all the readiness variables.

henrjk (commented):
Looks good to me except typo (see extra comment)

common/ocp-config/component/template.yml
michaelsauter: Oh my :( Thanks for the catch! Fixed in 949fc12.

gerardcl (approved):
LGTM! 🚀

gerardcl:
this means we can start/continue now on adding the missing probes into the quickstarters! great!!

PR #479, opened 2020-10-27
Use @shared_library_ref@ and @agent_image_tag@ in Jenkinsfile templates

michaelsauter:
Follow-up from https://github.com/opendevstack/ods-jenkins-shared-library/pull/457.

### DIFF
--- a/be-fe-mono-repo-plain/Jenkinsfile.template
+++ b/be-fe-mono-repo-plain/Jenkinsfile.template
@@ -1,10 +1,10 @@
 // See https://www.opendevstack.org/ods-documentation/ for usage and customization.
 
-@Library('ods-jenkins-shared-library@@ods_git_ref@') _
+@Library('ods-jenkins-shared-library@@shared_library_ref@') _
 
 // depending on what techs are needed, a different agent image may be used below
 odsComponentPipeline(
-  imageStreamTag: '@ods_namespace@/jenkins-agent-base:@ods_image_tag@',
+  imageStreamTag: '@ods_namespace@/jenkins-agent-base:@agent_image_tag@',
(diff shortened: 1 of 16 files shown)


be-fe-mono-repo-plain/Jenkinsfile.template
```diff
@@ -1,10 +1,10 @@
 // See https://www.opendevstack.org/ods-documentation/ for usage and customization.
```
gerardcl, line 1: Maybe we could add in the headers of the templates the explanation you documented (link to it) about the relation of the different version tags. So everyone can know why a specific version number is there.

michaelsauter: @gerardcl I'm not sure I understand what you want to do ... could you give an example?

clemensutschig: @michaelsauter 
https://www.opendevstack.org/ods-documentation/opendevstack/3.x/quickstarters/be-golang-plain.html 

so render the tag into it (in this case 3.x) - so people can just copy this line - and are directly in the right doc

michaelsauter: Hmm, but:

1. over time people update the Jenkinsfile (when they migrate to the next version), so the link gets outdated ... or they also need to maintain the comment ...
2. admins might use e.g. `ODS_IMAGE_TAG=3.0.x`, in which case the generated link would be wrong from the start

michaelsauter:
OK to merge @clemensutschig @gerardcl? I'm not certain about the documentation change, and consider it a separate topic anyway.

clemensutschig (approved):
Go ..

PR #480, opened 2020-11-02
Fixed bug in multi-line training scripts

henninglebbaeus:
### This PR concerns the ds_ml_service

I encountered an issue when training on a remote machine. The problem is that the mulit-line scripts in `/ods-quickstarters/ds-ml-service/files/src/mlservice/services/infrastructure/remote/ssh/executors.py`  is broken, such that only the first statement is executed. This had the effect that command line parameters were not passed to the training script.

I fixed the scripts in the `run_training` method by adding a **\\\\** add the end of each line.

### DIFF
--- a/ds-ml-service/files/src/mlservice/services/infrastructure/remote/ssh/executors.py
+++ b/ds-ml-service/files/src/mlservice/services/infrastructure/remote/ssh/executors.py
@@ -135,13 +135,18 @@ mkdir {0}""".strip().format(self._target_folder),
         """
         # remote training with remote dvc repository
         if self._dvc_remote:
-            train_script = """{0}/services/infrastructure/remote/scripts/remote_trainer.py --env {0}
-            --debug {1} --dvc_remote {2} --dvc_user {3} --dvc_password {4}""". \
+            train_script = """{0}/services/infrastructure/remote/scripts/remote_trainer.py \\
+                --env {0} \\
+                --debug {1} \\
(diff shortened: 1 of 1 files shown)


ds-ml-service/files/src/mlservice/services/infrastructure/remote/ssh/executors.py
```diff
@@ -135,13 +135,18 @@ def run_training(self) -> None:
         """
         # remote training with remote dvc repository
         if self._dvc_remote:
-            train_script = """{0}/services/infrastructure/remote/scripts/remote_trainer.py --env {0}
-            --debug {1} --dvc_remote {2} --dvc_user {3} --dvc_password {4}""". \
+            train_script = """{0}/services/infrastructure/remote/scripts/remote_trainer.py \\
+                --env {0} \\
```
gerardcl, line 139: hi @henninglebbaeus , 
did you tried exactly the same structure but only one backslach?

henninglebbaeus: I think it has to be a double backslash, since backslash is an escape character in Python.

gerardcl: got it! basically we need to escape it! thanks!

gerardcl (approved):

Issue #483, opened 2020-11-09
Transition quickstarters from NodeJS 10 to NodeJS 12


michaelsauter:
We already have a NodeJS 12 agent image, but it is not installed by default, and not used by any quickstarter, see https://github.com/opendevstack/ods-quickstarters/issues/469. Also, the NodeJS 12 image does not contain necessary dependencies for E2E tests (e.g. with Cypress).

For ODS 4, I suggest we:

* remove the `nodejs10-angular` agent image
* add the required dependencies (chrome, xvfb) to the `nodejs12` agent images
* switch all quickstarters to use `nodejs12`

As a consequence, users who want to adopt ODS 4 also need to move from NodeJS 10 to 12.

Why 12 and not 14? RedHat also only offers 12 (https://github.com/openshift/jenkins/tree/master/agent-nodejs-12), so I think we should stick with this.

Objections @metmajer @clemensutschig @cschweikert @ManuelFeller @m-apsolon?

m-apsolon:
sounds like an excellent idea to me.

clemensutschig:
doc impact & test impact (check for the right agent image)

PR #485, opened 2020-11-13
Transition from nodejs10-angular to nodejs12-angular

felipecruz91:
Transition from nodejs10-angular to nodejs12-angular.

Refers #483 

Tasks: 
- [x] remove the nodejs10-angular agent image
- [x] add the required dependencies (chrome, xvfb) to the nodejs12 agent images
- [x] switch all quickstarters to use nodejs12

Let me know if there's anything more that I should include in this PR.


michaelsauter (changes requested):
Awesome, thanks for this!

One request:

Can you drop the `-angular` from the new image? I think this is highly misleading. It should be `jenkins-agent-nodejs12` only. However, it should include chrome/xvfb like you have it right now.

And one comment:

I'm currently reworking how these deps are installed when the base image is a UBI8. Instead of downloading a single package from a mirror, it configures the repo properly. I think this might be worthwhile to backport to the non-UBI8 as well, but we can do this after your PR is merged.

felipecruz91:
@michaelsauter The pipeline job `jenkins-agent-nodejs12` will fail because it uses a fake nexus URL when building the Docker image:  `--build-arg nexusUrl=https://nexus.example.com`.

common/jenkins-agents/nodejs12/docker/Dockerfile
```diff
@@ -1,52 +1,87 @@
...
+
+RUN npm config set registry=$nexusUrl/repository/npmjs/ && \
+    npm config set always-auth=true && \
+    npm config set _auth=$(echo -n $nexusAuth | base64) && \
+    npm config set email=no-reply@opendevstack.org && \
+    npm config set ca=null && \
+    npm config set strict-ssl=false && \
+    yarn config set registry $nexusUrl/repository/npmjs/ -g && \
+    npm install -g @angular/cli@8.0.1 --unsafe-perm=true --allow-root && \
+    npm install -g cypress@3.3.1 --unsafe-perm=true --allow-root > /dev/null && \
```
michaelsauter, line 79: @cschweikert can we drop installing `@angular/cli@8.0.1` and `cypress@3.3.1`? That would be ideal because we can then get GitHub Actions to work. Further, it feels wrong for me to back versions of these tools into the agent.

cschweikert: * ``@angular/cli@8.0.1``: We can try to remove it. It is installed now via Jenkinsfile during provisioning: https://github.com/opendevstack/ods-quickstarters/blob/2c4514d411c7cabedda191f94fcf6a0798530e37/fe-angular/Jenkinsfile#L13
* ``cypress@3.3.1``: Having this in was nice, when a build should run e2e tests without prior installation of the cypress testing framework. This usually takes about one minute. But then again you are limited to this specific version. There are other ways of baking cypress binaries into a custom agent image. Last time I did it with a ``Dockerfile`` like this:

```
# This FROM tag is informational. It is overwritten by OpenShift (see oc-template.yaml file)
FROM cd/jenkins-slave-nodejs10-angular:2.x

USER root

# add cypress@4.5.0 binaries to /home/jenkins/.cache/Cypress/4.5.0/Cypress/...
# this allows to disable post-installation of cypress@4.5.0 by setting the environment variable
#   CYPRESS_INSTALL_BINARY=0
# cypress@4.5.0 can use those binaries by setting environment variable to the cache folder that holds the binaries
# for the respective versions of cypress
#   CYPRESS_CACHE_FOLDER=/home/jenkins/.cache/Cypress
ADD https://download.cypress.io/desktop/4.5.0?platform=linux&arch=x64 /home/jenkins/cypress/cypress-linux.zip
RUN md5sum /home/jenkins/cypress/cypress-linux.zip && \
    echo "26835296a4edb1dea35795f59213c519 /home/jenkins/cypress/cypress-linux.zip" | md5sum -c && \
    mkdir -p /home/jenkins/.cache/Cypress/4.5.0 && \
    unzip -q /home/jenkins/cypress/cypress-linux.zip -d /home/jenkins/.cache/Cypress/4.5.0

# fix access rights
RUN chgrp -R 0 $HOME && \
    chmod -R g=u $HOME && \
    chmod ug=rx /home/jenkins/.cache/Cypress/4.5.0/Cypress/Cypress
USER 1001
```

This will allow skipping the post-installation step of cypress binaries (faster build, no issues with proxy servers, adds missing file consistancy check (MD5) for the binaries, etc.)

tl;dr

Yes, let's drop installing ``@angular/cli@8.0.1`` and ``cypress@3.3.1``.

michaelsauter: Cool. Yes I'm in favour of that approach. We should have a section in the docs with "recipes" like this one.

docs/modules/jenkins-agents/partials/update-to-3x.adoc
```diff
@@ -1,6 +1,6 @@
...
+Some agent images (`jenkins-agent-maven`, `2-angular`, `jenkins-agent-scala`) exposed the environment variables `NEXUS_HOST` and `NEXUS_USERNAME`/`NEXUS_PASSWORD`/`NEXUS_AUTH`. This was not done intentionally, but as a side effect of using those variables during image build time. In 3.x, that has been cleaned up. Now those variables are passed as build arguments to the image build, which means they are no longer accessible as environment variables in the running container. Nexus credentials can and should be accessed from the pipeline context (`context.nexusUrl`, `context.nexusUsername` and `context.nexusPassword`). As a consequence, if a `Jenkinsfile` relies on the presence of those environment variables without setting them explicitly using `withEnv`, it will fail to run in 3.x. That said, the `Jenkinsfile` templates which shipped by default with the quickstarters did not rely on the environment variables - with the exception of `fe-ionic` which did not set `NEXUS_HOST` explicitly in the `Build` stage.
```
michaelsauter, line 3: `2-angular` seems like a mistake. Maybe some replacement gone wrong?

I think this file should not change at all - it's an update from 2.x to 3.x, which should not be affected by new developments.

michaelsauter (changes requested):
Almost 👍 but maybe you can fix the one doc issue I found before we merge? Thanks!

docs/modules/jenkins-agents/partials/update-to-3x.adoc
felipecruz91: That's true - it was a mistake when replacing all occurrences. Let me put it back.

michaelsauter (approved):

PR #489, opened 2020-11-16
Add new Dockerfile.ubi8 for jenkins-agent-python

faust2199:
Tested with Michael's most recent base image.


common/jenkins-agents/python/docker/Dockerfile.rhel8
```diff
@@ -0,0 +1,29 @@
+FROM opendevstackorg/ods-jenkins-agent-base-centos7:latest
```
michaelsauter, line 1: I guess we should wait with merging this until we have published the new UBI8-based base image so that we can also add this image to the test suite.

michaelsauter: You can use https://hub.docker.com/layers/127012478/opendevstackorg/ods-jenkins-agent-base-ubi8/latest/images/sha256-8a0d86c6bff3758e3726b9722c4209dd4de948efdb48915cbc333267c6912e4c?context=explore now.

faust2199: Updated image reference and Dockerfile suffix.

michaelsauter:
@faust2199 Cool. Can you add a test for it like it is done in https://github.com/opendevstack/ods-quickstarters/blob/master/.github/workflows/continuous-integration-workflow.yml#L50?

michaelsauter (approved):

michaelsauter (changes requested):
Actually, can you rename the old one to `Dockerfile.rhel7`, and make it the default in the `BuildConfig`?

faust2199:
> Actually, can you rename the old one to `Dockerfile.rhel7`, and make it the default in the `BuildConfig`?

Sure. By the way, it passed the test with the log filled with errors about being unable to access various "*rhel8*" repos. Does it make sense to disable these repos in the base image so that only centos8 repos are used?

michaelsauter:
@faust2199 See also my changes in PR #494.

michaelsauter:
@faust2199 Interesting question. Not sure yet. Given those errors:

> Couldn't resolve host name for http://base-4-7-rhel8.ocp.svc/rhel-8-baseos/repodata/repomd.xml [Could not resolve host: base-4-7-rhel8.ocp.svc

It looks like it assumes that it is being built inside an OpenShift cluster. Not sure, but maybe it is possible to tell the build that it is running outside?

michaelsauter (approved):

michaelsauter (approved):

michaelsauter:
@faust2199 Can you please backport this to the 3.x branch as well? Thanks!

michaelsauter:
@faust2199 ping :)

PR #495, opened 2020-11-25
Add NodeJS 12 image based on UBI8

michaelsauter:

PR #497, opened 2020-11-26
Add NodeJS10 UBI8 agent image

michaelsauter:
This uses the work done in #495, but installs NodeJS 10, because that's the version used in 3.x. We'll also backport #495 once merged.

There is no GitHub Action for this unfortunately because we install NPM packages for which we would need a running Nexus. I tested the image against the `fe-angular` quickstarter though. The new NodeJS 12 image (#495) does not install packages and therefore is tested in GitHub.

### DIFF
--- a/common/jenkins-agents/nodejs10-angular/docker/Dockerfile.rhel7
+++ b/common/jenkins-agents/nodejs10-angular/docker/Dockerfile.rhel7
@@ -15,7 +15,7 @@ LABEL com.redhat.component="jenkins-agent-nodejs-rhel7-docker" \
 ARG nexusUrl
 ARG nexusAuth
 
-ENV NODEJS_VERSION=8.11 \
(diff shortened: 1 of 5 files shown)


common/jenkins-agents/nodejs10-angular/docker/Dockerfile.rhel7
```diff
@@ -15,7 +15,7 @@ LABEL com.redhat.component="jenkins-agent-nodejs-rhel7-docker" \
 ARG nexusUrl
 ARG nexusAuth
 
-ENV NODEJS_VERSION=8.11 \
```
michaelsauter, line 18: This is not actually used, but I thought it makes sense to "fix" it. The variable is used again in the new NodeJS 12 image so I left it here.

faust2199 (approved):

PR #504, opened 2020-11-30
Add UBI8 Maven agent image

michaelsauter:
Tested with the `be-java-springboot` quickstarter.


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -1,210 +1,120 @@
...
+#       uses: actions/checkout@v2.0.0
+#     -
+#       name: Build docker image
+#       working-directory: common/jenkins-agents/scala/docker
+#       run: |
+#           docker build --tag agent-scala-test \
+#             --build-arg nexusUrl=https://nexus.example.com \
+#             --build-arg nexusUsername=developer \
+#             --build-arg nexusPassword=s3cr3t \
+#             .
```
michaelsauter, line 120: Changes due to CRLF -> LF, and then I used that opportunity to also removed the commented `test` job which is no longer in use, and unlikely to be in use anytime soon.

oalyman (approved):

PR #513, opened 2020-12-02
Fix java continue on unit test errors

hrcornejo:
Fix a problem returnig failed unit tests to Jira

### DIFF
--- a/be-java-springboot/templates/build-4.10.gradle
+++ b/be-java-springboot/templates/build-4.10.gradle
@@ -28,6 +28,8 @@ plugins {
 }
 
 ext['spring-restdocs.version'] = '2.0.2.RELEASE'
+// When a test fail we can continue or fail the stage
+ext['CONTINUE_WHEN_TEST_FAIL'] = true
 
 apply plugin: 'io.spring.dependency-management'
 
@@ -76,6 +78,7 @@ test {
     outputs.dir snippetsDir
     useJUnitPlatform()
     finalizedBy jacocoTestReport
+    ignoreFailures = "${CONTINUE_WHEN_TEST_FAIL}"
 }
 
 asciidoctor {


stitakis (changes requested):
We should hold on a bit on this change and investigate if this hardcoded value could be passed from the jenkins shared lib. Without this change, gradle fails fast. Is not was we need for the release manager. What we need here is to avoid stopping the gradle build in case a test fails.

michaelsauter:
Also note that If this change is motivated by some incorrect behaviour in the orchestration pipeline, then changing the quickstarter will not really fix the problem, at least not for any existing components: there is no connection between the quickstarter and its provisioned component.

@stitakis A component pipeline can know if it has been triggered by the orchestration pipeline, so in theory we can behave differently in that case. However, I would not like to exploit that extensively as it would increase the effort needed to go to a design in which the orchestration pipeline would only collect artefacts, but not create any (which I think would be very attractive, but I have not looked into feasibility yet).

stitakis:
@michaelsauter yes, I agree on this:
> Also note that If this change is motivated by some incorrect behaviour in the orchestration pipeline, then changing the quickstarter will not really fix the problem

regarding
> A component pipeline can know if it has been triggered by the orchestration pipeline...
can you please point me to the specific code? Where the orchestration pipeline calls the component pipeline? How does the component pipeline know that it has been triggered to by the orchestration pipeline

hrcornejo:
Some additional info. This is the execution without changes in quickstarter:

image

And this is the execution changing the source code:

image

The WIP version does not continue with later stages (generate doc, inform Jira, deploy....) when de Build phase is wrong.

hrcornejo:
I don't lnow if other solution could be done here: https://github.com/opendevstack/ods-jenkins-shared-library/blob/3.x/vars/odsOrchestrationPipeline.groovy#L96 or here: https://github.com/opendevstack/ods-jenkins-shared-library/blob/3.x/src/org/ods/orchestration/BuildStage.groovy#L21 but the most important is to clarify the behaviour. With a failed unit test the rest of stages must continue or not?

stitakis:
@metmajer @michaelsauter 
the spring boot quickstarter (and I guess the other ones too) is configured by default to fail fast. This means, if the unit test stage fails (and the same for linting stage), then its pipeline fails (first screenshot above). This is the current behavior for component pipelines and make sense during development phase.

But, when preparing a release (e.g. developer preview) it looks to me that the expected behavior should be different (see 2nd screenshot) than the current implementation. In this case the orchestration pipeline could run to the end stage even if the unit test fails. While I see a value for the user on this, the problem is that the current implementation does not support it. Another problem related to this is that if the orchestration pipeline doesn't run to the end, then the failed unit tests are not reported in jira as bugs.

To change this situation:
A: one solution to explore would be to just only modify the orchestration pipeline to ignore a failure in the component pipeline.
B: we could try to report to jira the bug earlier in the orchestration pipeline, just after a component pipeline fails (and/or additionally let the orchestration pipeline fails if one of the component pipeline failed)
C: we could modify the spring boot quickstarters (and also all others) to ignore failures in unit test when the component pipeline is called from the orchestration pipeline.
D: we could leave as it is, so that every component pipeline needs to run successful in order the orchestration pipeline to succeed.

Apart from D, all other solutions would take some time. I'm not sure if we'll make before the release of ODS 3.1.0.

How we should continue on this? Your thoughts?

michaelsauter:
@stitakis Thanks for the detailed write-up.

My suggestion is to stick with option D for now. Changing this has too much impact for 3.1 in my eyes.

For ODS 4, we should discuss this, but I think we should proceed with caution. A failed stage might indicate that further stages do not work either, or actually should not be executed. I'm not sure we really want to enforce a "continue on failure anyway".

Maybe a "failing" component pipeline is OK but the orchestration pipeline needs to change how it deals with this case? Is the problem we are trying to solve "just" that Jira is not informed of failing unit tests?

stitakis:
@michaelsauter @metmajer yes, the problem we are trying to solve is that Jira is not informed of failing unit test. I also think that for 3.1 we should stick to option D to avoid any regression problem.

michaelsauter:
Then my proposal is to look into reporting failing unit tests to Jira also in the case of a failed pipeline. This should be the responsibility of the orchestration pipeline, so the fix should be there (if possible, I have not looked into feasibility yet).

stitakis:
@hrcornejo @metmajer I agree with Michaels proposal... let us close this PR and open an issue in ods-quickstarter project, ok?

michaelsauter:
Issue should be in `ods-jenkins-shared-library` I think ...

stitakis:
yes, it should be in `ods-jenkins-shared-lib` -> @hrcornejo @metmajer do you agree to the last proposal?

hrcornejo:
@stitakis ok for me

metmajer:
@stitakis @michaelsauter @hrcornejo if we do not enable tests to continue upon failure, the pipeline will stop when it sees the first test failure. Fast feedback is our all goal, but there is little value in providing feedback only for a single test with every pipeline run. The current behavior has the side effect that, in the EDP, all tests after the first failing tests are reported to be missing from the execution, which is ambiguous. What's the impact when changing the current behavior? I do not see how the jenkins shared library could provide the intended behavior.

michaelsauter:
I think the test runner will execute all tests, regardless whether there are failing tests or not. This is the current behaviour as far as I am aware of (I would not know of a test runner which stops after the first failing test by default).

After the "test step" fails, the component pipeline should fail as well I think. Again this is the current behaviour. The orchestration pipeline would not need to fail though I think - and to me it sounds like currently it fails and reports everything else as missing. That seems wrong and should be fixed. I think the right place to do this is the orchestration pipeline. Could it be that `failFast` (https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/orchestration/BuildStage.groovy#L73) is the issue here?

metmajer:
> I think the test runner will execute all tests, regardless whether there are failing tests or not. This is the current behaviour as far as I am aware of (I would not know of a test runner which stops after the first failing test by default).
> 
> After the "test step" fails, the component pipeline should fail as well I think. Again this is the current behaviour. The orchestration pipeline would not need to fail though I think - and to me it sounds like currently it fails and reports everything else as missing. That seems wrong and should be fixed. I think the right place to do this is the orchestration pipeline. Could it be that `failFast` (https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/orchestration/BuildStage.groovy#L73) is the issue here?

The `failFast` fails fast if there is a failure in one of the components. The ods-test component type, however, has been designed to fail only once all tests have executed and a complete overview of which tests have failed is available. Therefore, all tests of type `integration`, `acceptance`, `installation` will be correctly reported. When failing tests from an ods-test component are known the Release Manager will not fail the pipeline immediately but will mark the pipeline failed at the very end in StageFinalize.

The Spring Boot quickstarter, for example, returns an error after the first failing unit test, which differs from the above-mentioned behavior. Here, the Release Manager pipeline will fail immediately, since the execution of unit tests is regarded in the context of the "build" of the component. Enabling the execution of all tests in quickstarters is IMHO the best approach we have to cover all feedback. I am wondering though why the component advances past the build stage although there has been a failing test. @hrcornejo could it be that the pipeline issue was because of a different reason? Can you provide a log please?

hrcornejo:
@metmajer 
[log - Failed.txt](https://github.com/opendevstack/ods-quickstarters/files/5694164/log.-.Failed.txt)
The log of build 2 (same as screenshot). And build 3: 
[log.txt](https://github.com/opendevstack/ods-quickstarters/files/5694184/log.txt)

metmajer:
@hrcornejo Hector, can you please confirm that by not using `ignoreFailures = true` the unit tests do not get executed? If they get executed, then the error in the feedback of information to Jira is not in the component, but in the Release Manager.

hrcornejo:
@metmajer without `ignoreFailures` the tests ARE executed and the build marked as failed:
`3 tests completed, 1 failed

> Task :test FAILED
> Task :jacocoTestReport

FAILURE: Build failed with an exception.`

With the `ignoreFailures=true` the tests are executed but the build is not marked as failed:
`3 tests completed, 1 failed
There were failing tests. See the report at: file:///tmp/workspace/edpt3-cd/edpt3-cd-demo-app-release-manager-master/build/reports/tests/test/index.html

> Task :jacocoTestReport
> Task :check
> Task :build

BUILD SUCCESSFUL in 3m 0s`

metmajer:
@hrcornejo what is the status code of the `gradlew test` command execution? Is it != 0 in both cases?

hrcornejo:
No, is the reason this fails in Jenkinsfile:
[...]
def status = sh(script: "./gradlew clean build --stacktrace --no-daemon", returnStatus: true)
      if (status != 0) {
        error "Build failed!"
      }
[...]

stitakis:
> @hrcornejo concerning the bug you are trying to fix, I would consider the reported behavior to be OK for now.

@metmajer did I get it right? Are we descoping this from 3.1 release?

metmajer:
@hrcornejo after looking at the code again... if you say that all tests are executed, then we can restore the wanted behavior simply by removing `group.failFast = true` in `BuildStage.groovy`. Can you pleaes give this a try?

metmajer:
> @metmajer no, is the reason this fails in Jenkinsfile (0 ->success):
> [...]
> if (status != 0) {
> error "Build failed!"
> }
> [...]

I believe that we can overcome the `error` you have outlined. If you look at `MROPipelineUtil.prepareExecutePhaseForRepoNamedJob` you will find that ODS component pipelines are executed in calls to `executeODSComponent`, which are wrapped in a block `executeBlockAndFailBuild`, which looks as follows:

```
    void executeBlockAndFailBuild(Closure block) {
        try {
            block()
        } catch (e) {
            this.failBuild(e.message)
            hudson.Functions.printThrowable(e)
            throw e
        }
    }
```

If we catch the `throw e` inside `StageBuild`, we should be able to avoid the pipeline to stop completely. The pipeline will be marked red, but control flow will continue. If we then stop the pipeline after we reported test results into Jira, we have what we need. Example code:

```
try {
    execute build stage
} catch (e) {
    // error in build has happened
    if (hasFailingTests) {
        reportFailingTests()
    }

    throw e
}
```

Thoughts @hrcornejo @stitakis?

metmajer:
> @metmajer did I get it right? Are we descoping this from 3.1 release?

@stitakis not if we can find a simple solution :)

metmajer:
Adding the desired handling is not as simple as I thought and will make for an ugly hack. We'll have to defer. Stopping the pipeline after every unit test is a bit of a pain, but we cannot come up with a better solution at this point in time.

metmajer:
I am closing this issue. We yet need to figure out how to move forward in the near future.

clemensutschig:
Can someone please Paste the stack and log from the release Manager..  ? Something changed i guess... (jenkins?) .. the default behavior is to set the build to fail but move on... See pipeline groovy Component Pipeline
.  @hrcornejo @metmajer

metmajer:
@clemensutschig the desired behaviour is as follows:
1) when the build stage has errors, the orchestration pipeline fails
2) my expectation is: the build stage executes completely (runs all tests); pipeline fails only at end of stage (not after a single failing unit test). rationale: there could be too many failing ones and since they execute fast we want the full feedback
3) all unit tests are executed; when there have been failing unit test, the test results are reported into Jira

PR #514, opened 2020-12-03
Read more environment variables for the QS pipeline

michaelsauter:
* Read `SHARED_LIBRARY_REF` and `AGENT_IMAGE_TAG` if present. This is a
follow-up from https://github.com/opendevstack/ods-core/issues/773. Note
that the default Git ref changes from `master` to `latest`, but this
should be fine as the ODS installation creates this Git ref. The
component `Jenkinsfile` already points to `latest` by default.
* Read environment variables within `node` block to also pick up
defaults baked into the Jenkins master image.

Ran into this during testing for OpenShift 4 and also in tests related to https://github.com/opendevstack/ods-provisioning-app/issues/646.


clemensutschig (approved):

gerardcl (approved):
🚀

oalyman (approved):

PR #516, opened 2020-12-07
Fix quickstarters

michaelsauter:
https://github.com/opendevstack/ods-jenkins-shared-library/pull/514 had more impact on the quickstarters than expected:

* `ds-rshiny` and `ds-jupyter-notebook` still worked, but they were rolled out twice, which is not necessary.
* `be-fe-mono-repo-plain` used the rollout information to check the deployment. That failed because the structure changed.
* `ds-ml-service` is using a "staged approach": first build and deploy one service, then continue with the next. Now that we everything in one go that has a certain label, we need to set and target more specific labels to achieve this.

Fixes https://github.com/opendevstack/ods-jenkins-shared-library/issues/526.

### DIFF
--- a/be-fe-mono-repo-plain/Jenkinsfile.template
+++ b/be-fe-mono-repo-plain/Jenkinsfile.template
@@ -57,10 +57,16 @@ def stageUnitTestBackend(def context) {
 }
 
 def stageTestDeployedComponents(def context, def deploymentInfo) {
+  def podName = deploymentInfo['DeploymentConfig'][context.componentId][0].podName
   stage('Test Deployed Component') {
-	sh (script : "oc port-forward ${deploymentInfo.podName} 8081 -n ${context.targetProject} &", label : "Setting up port forward for pod ${deploymentInfo.podName}")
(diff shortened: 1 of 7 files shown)


clemensutschig:
Looks good to me.. can you also Check the docs tip/ tir for those changes... are the components there and cleany propagate from dev to q

michaelsauter:
@clemensutschig I will run more tests with the release manager anyway once all the changes are in. Just trying to break things down into small reviewable pieces. 

That said, the release manager should not be affected as the communication happens via the artefacts, see https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/component/RolloutOpenShiftDeploymentStage.groovy#L133-L134.

gerardcl:
in this case it was fine to remove the oauth proxy deployments since we always proxy forward to the service name instead of using injected env. vars. 👍

michaelsauter:
@gerardcl The stage now also takes care of both deployments (also the auth proxy) - so there should not be any change actually to the previous version

PR #520, opened 2020-12-08
Maven Agent- Select Java 11 version explicitly (master)

michaelsauter:
Try to fix build failures since yesterday ... we also used the same approach in RHEL7.


clemensutschig:
@michaelsauter do you want Merge this prior noon the next ci run?

clemensutschig (approved):

michaelsauter:
@clemensutschig Merged as it makes the GitHub build green, but it is not used in the scheduled CI run (that uses OpenShift 3.11 and therefore the RHEL7 images).

PR #522, opened 2020-12-11
Added assets to run quickstarters tests with Docker

felipecruz91:
The following is an improvement I'd like to suggest to streamline the process of getting any cluster administrator to run the quickstarter tests.

As a cluster administrator, I need to spend some time configuring my local machine with the tools required to run the quickstarter tests. As of today, some of those tools are _golang_, _go-junit-report_, _oc_ and _jq_. I believe it would be beneficial that OpenDevStack provides an official Docker image that cluster administrators can just simply pull and run the quickstarter tests 😀

For instance:

```cli
cd $UMBRELLA_DIR

docker run --rm -it \
    -v $(pwd):/work \
    opendevstack/ods-quickstarters-tests-runner
```

I would like to propose the changes in this PR to make it happen if you consider it appropriate.

Tasks: 


metmajer:
@felipecruz91 would it make sense to hide the complexity of setting up the docker container images behind the `make test` or a `make test-docker` target?

felipecruz91:
@metmajer Thanks for your valuable input! I've extended the Makefile with the `test-docker` target.

tests/ods-quickstarters-tests-runner/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+# ENV http_proxy <SET_VALUE> 
+# ENV https_proxy <SET_VALUE>
+# ENV HTTP_PROXY <SET_VALUE> 
+# ENV HTTPS_PROXY <SET_VALUE>
+
+RUN go get -u github.com/jstemmer/go-junit-report
+
+RUN wget https://github.com/openshift/origin/releases/download/v3.11.0/openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit.tar.gz && \
+    tar -xzf openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit.tar.gz && \
+    cd openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit && \
```
michaelsauter, line 13: This is hardcoding the `oc` version to 3.11. Not ideal if you want to run against an OpenShift 4 cluster, which is a valid use case right now.

michaelsauter (commented):
Thanks for sharing this!

TBH, I'm a bit torn on this. I am not sure we actually make it easier to run those tests:

* Previously you need `oc`, `jq` and Go. I think `oc` is a given if you do anything with ODS, and `jq` is something you should have on your machine anyway (IMHO). So the only new thing which people might not have on their machine is Go
* With this PR, I need `docker`, and I need to take extra care about proxies. Having `docker` installed is not needed for ODS in general ... so I feel we are weighing installing Docker vs. installing Go.

So with this in mind I feel like this might add more to maintain and setup than you need right now, plus might lead to bringing in even further deps because they can be hidden in the image.

What are your thoughts? Views of others?

michaelsauter:
Oh just noticed this in `ods-quickstarters`, and against the `3.x` branch. The test framework has been moved to `ods-core`. If we were to go ahead with this, the changes should be done there, and against `master`.

metmajer:
@michaelsauter @felipecruz91 I share Michael's thoughts to some degree, but I think this enhancement adds some extra convenience for those who have docker available, and don't care much about the rest. I would see this addition as more critical if it would replace the pre-existing approach. Also, Felipe's addition to the documentation states that you "can use docker". I would be fine with it. Soon, we would like to bring the task of managing ODS installations to our developers. The fewer steps to set up a local environment the better IMHO.

felipecruz91:
@michaelsauter Thanks for the input. I understand your concerns about having to maintain another way to run the quickstarter tests (installing Docker vs installing Go).

My vision is to provide an _additional_, almost-immediate way to run the tests to avoid developers spending time preparing their local environment. So I thought that if ODS provided a Docker image, this would be very convenient for developers with Docker _already_ installed; they could benefit from this by running the tests with a simple `docker run` command.

Nevertheless, if you reckon this will difficult the maintainability compared to the benefits developers could get from it, then I totally respect that :)

michaelsauter:
> My vision is to provide an additional, almost-immediate way to run the tests to avoid developers spending time preparing their local environment. So I thought that if ODS offered a public Docker image in DockerHub, this would be very convenient for developers with Docker already installed; they could benefit from this by running the tests with a simple docker run command.

I think a public image won't fully work because a) you're hard coding the OpenShift version and b) you can't know the proxy settings.

Therefore I personally would rather invest into a make target which checks the prerequisites and gives clear hints what to do if something is missing. Something like:

```
check-env:
  which jq || echo 'jq is missing, install from https://stedolan.github.io/jq/'
  ...
```

However, I don't want to be in the way of this ... just voicing my concerns that maintaining this feels more work than it is worth.

clemensutschig:
for me this seems some un-rewarded complexity... (also in dealing with proxies etc).
as admin - go clone the ods-core / quickstarters and configuration and run make test - done .. 

I rather bake a small install script to install go / and jq and oc - and that's it 👍

felipecruz91:
No probs at all, thanks for sharing your thoughts! 😀

PR #528, opened 2020-12-16
Make go distribution URL configurable and use embedded golangci-lint installer

### LINKED ISSUE #524: Make golang agent buildable on OpenShift installations running in mainland China
Due to network restrictions, the golang agent fails to build on OpenShift installations running in mainland China. Two issues are blocking the build. Their descriptions and possible solutions are listed below:

1. Unable to download golang binaries from https://storage.googleapis.com/golang/go$GO_VERSION.linux-amd64.tar.gz

- Possible solution:
  - Optionally use alternative mirrors to download binaries. Recommendation from golang.org: https://blog.golang.org/hello-china

2. Unable to download golangci-lint installer from https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh

- Possible solutions:
  - Embed the installer script alongside the Dockerfile, so that we can just COPY it to the container instead of downloading it from the Internet. The installer is version agnostic and only changed infrequently (last updated 8 months ago).
  - Obtain the file by "git clone" the repo. The restriction only applies to raw.githubusercontent.com. Other functionalities of GitHub are normal.

faust2199:
Closes #524


michaelsauter (approved):
Cool 👍 

As this updates Go from 1.14 to 1.15, did you check if there is any other mentioned / reference to Go 1.14, e.g. docs?

faust2199:
> Cool 👍
> 
> As this updates Go from 1.14 to 1.15, did you check if there is any other mention / reference to Go 1.14, e.g. docs?

Just checked. Two files contain references to Go 1.14:

ods-quickstarters\CHANGELOG.md
ods-quickstarters\docs\modules\jenkins-agents\pages\golang.adoc

It appears these files are updated when a new major version is released.

Another thing probably worth mentioning:
- Our current way of installing golangci-lint is probably from https://golangci-lint.run/usage/install/#ci-installation, where the recommended command line has a specific version tag at the very end. However, according to the installer script, the version tag is optional. If the version tag is not specified, it will automatically find the latest release version and install that version. I guess it is a nice way of automatically keeping the version up-to-date. Source: https://github.com/golangci/golangci-lint/blob/e26630a82cc444935c571069bf4d5af7ae2e20e7/install.sh#L106

michaelsauter:
> Just checked. Two files contain references to Go 1.14:
>
> ods-quickstarters\CHANGELOG.md
> ods-quickstarters\docs\modules\jenkins-agents\pages\golang.adoc
>
> It appears these files are updated when a new major version is released.

Changelog yes, but the Asciidoctor file should be updated as we go to avoid we miss it later on. Could you update it?

> Another thing probably worth mentioning:
> 
> Our current way of installing golangci-lint is probably from https://golangci-lint.run/usage/install/#ci-installation, where the recommended command line has a specific version tag at the very end. However, according to the installer script, the version tag is optional. If the version tag is not specified, it will automatically find the latest release version and install that version. I guess it is a nice way of automatically keeping the version up-to-date. Source: https://github.com/golangci/golangci-lint/blob/e26630a82cc444935c571069bf4d5af7ae2e20e7/install.sh#L106

I guess this is a bit of an endless debate. Both approaches have pros and cons. Given the context ODS is used in, it is important we pin versions as much as possible to avoid build failures, and have transparency on what actually gets installed (= same versions in QA as in prod).

faust2199:
> > Just checked. Two files contain references to Go 1.14:
> > ods-quickstarters\CHANGELOG.md
> > ods-quickstarters\docs\modules\jenkins-agents\pages\golang.adoc
> > It appears these files are updated when a new major version is released.
> 
> Changelog yes, but the Asciidoctor file should be updated as we go to avoid we miss it later on. Could you update it?
> 
> > Another thing probably worth mentioning:
> > Our current way of installing golangci-lint is probably from https://golangci-lint.run/usage/install/#ci-installation, where the recommended command line has a specific version tag at the very end. However, according to the installer script, the version tag is optional. If the version tag is not specified, it will automatically find the latest release version and install that version. I guess it is a nice way of automatically keeping the version up-to-date. Source: https://github.com/golangci/golangci-lint/blob/e26630a82cc444935c571069bf4d5af7ae2e20e7/install.sh#L106
> 
> I guess this is a bit of an endless debate. Both approaches have pros and cons. Given the context ODS is used in, it is important we pin versions as much as possible to avoid build failures, and have transparency on what actually gets installed (= same versions in QA as in prod).

I like the idea of pinning versions to make builds more reproducible. I don't have a compelling reason to update the version of Go either. I've changed the version back to 1.14.2.

michaelsauter (approved):

PR #530, opened 2020-12-17
Expect images in cd project

michaelsauter:


michaelsauter:
@clemensutschig @stitakis This makes `make test-quickstarter QS=ods-document-gen-svc` and `make test-quickstarter QS=ods-document-gen-svc` pass again on my machine. I must have made an error when I ran it on the PR that changed the build to be in the `*-cd` project because I was able to reproduce the problem now.

PR #533, opened 2021-01-06
Maven Agent- Select Java 11 version explicitly (3.x)

michaelsauter:
Apply https://github.com/opendevstack/ods-quickstarters/pull/520 for 3.x as well. This only affects the UBI8 image.


gerardcl (approved):
🚀

PR #535, opened 2021-01-13
fixes ods ami build branch 3.x failing due nodejs12 jenkins agent wrong dockerfile name

stitakis:
ODS AMI build for branch 3.x is failing due to a wrong filename of dockerfile referenced in jenkins nodejs12 agent.
"error: open /tmp/build/inputs/common/jenkins-agents/nodejs12/docker/Dockerfile: no such file or directory"

This change fixes this by setting the correct filename.

### DIFF
--- a/common/jenkins-agents/nodejs12/ocp-config/template.yaml
+++ b/common/jenkins-agents/nodejs12/ocp-config/template.yaml
@@ -18,6 +18,8 @@ parameters:
 - name: ODS_GIT_REF
   required: true
   value: production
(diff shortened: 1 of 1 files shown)


michaelsauter (approved):
FYI @metmajer @borja44 Note this has no effect on actual 3.x installations as they do not build this agent image. 3.x still uses NodeJS 10. The NodeJS 12 image was part of the code base when 3.x was cut, but not referenced by any quickstarters and/or installation instructions.

PR #515, opened 2020-12-04
Add AWS Terraform Quickstarter

nichtraunzer:
Here is the initial PR for the Terraform Jenkins agent and the AWS Terraform quickstarter.
Things we need to discuss: 
- name of quickstarter : inf-terraform-aws 
- name of agent : terraform (it will also support AZURE in the future)
- proper documentation

PR #565, opened 2021-04-13
Add AWS Quickstarter for 3.x

nichtraunzer:
see #515


braisvq1996:
Hi @metmajer , for this quickstarter it should be also necesary to update the Makefile in ods quickstarters and the Config map for the provisioning application in ods-core?

metmajer:
Yes, @braisvq1996, we need to follow up here. Do you want to provide some guidance to @nichtraunzer ?

borja44:
> Yes, @braisvq1996, we need to follow up here. Do you want to provide some guidance to @nichtraunzer ?

We're working on it

michaelsauter:
> update the Makefile in ods quickstarters and the Config map for the provisioning application in ods-core

Good point! To add on this: note that during an upgrade of an existing ODS installation, the list of quickstarters in the prov app does NOT get updated automatically. This is on purpose, to allow each installation to list their own quickstarters. As a consequence, you'll need to manually edit the prov app properties to add this quickstarter to an existing installation.

PR #567, opened 2021-04-15
Bump urllib by dependabot

nichtraunzer:
Accepted change from dependabot 3362c3d (see #566) and updated CHANGELOG

### DIFF
--- a/inf-terraform-aws/files/Pipfile.lock
+++ b/inf-terraform-aws/files/Pipfile.lock
@@ -1,148 +1,171 @@
 {
-    "_meta": {
-        "hash": {
-            "sha256": "aae4f27f548c96ade55497509700a4850f389c594ea88e789999f829dc624c21"
-        },
-        "pipfile-spec": 6,
-        "requires": {
-            "python_version": "3"
-        },
-        "sources": [
-            {
-                "name": "pypi",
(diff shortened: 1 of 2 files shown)


inf-terraform-aws/files/Pipfile.lock
```diff
@@ -1,163 +1,148 @@
...
+        "yq": {
+            "hashes": [
+                "sha256:1d2ad403504d306b5258b86c698f9856d7ad58b7bb17a2b875691a6a7b8c4c20",
+                "sha256:1f124f48dee77ad5e0be8607777fed183e96c8d31fa577de14201c7a614e4819"
+            ],
+            "index": "pypi",
+            "version": "==2.12.0"
+        }
+    },
+    "develop": {}
```
michaelsauter, line 147: Do you know why the list is shorter now? Seems like other reps got removed?

michaelsauter (approved):

inf-terraform-aws/files/Pipfile.lock
nichtraunzer: hm ... good catch ..   
I created the pipfile again in my environment and uploaded it .. both version have been tested successfully ... I might need to understand how dependabot did create the original one.
From my point of view  the latest commit is fine.

PR #569, opened 2021-04-15
ds-rshiny upgrade and housekeeping

### LINKED ISSUE #563: ds-rshiny Dockerfile update
The `ds-rshiny` is still using `ubuntu:16.04` as the base image. We need to upgrade it and port to a better managed base image.

To have a more recent OS base image (maybe UBI based) or to use one from the widely used https://github.com/rocker-org

To just upgrade the current ubuntu version.

**Additional context**
IMHO we should use base images that are not too much opinionated like maybe the rocker based ones are but just provide a really minimal baseline as we already do. Users will anyway tweak it a lot, it is R!

gerardcl:
I decided to focus on upgrading current offered base OS, so to update to Ubuntu 20.04.
I also did a first try with UBI8 and that will require a bit of more tweaks in terms of adding manually many dependencies required for having R in a proper shape. But it will be possible.

Fixes #563 

Tasks: 

### DIFF
--- a/ds-rshiny/files/docker/Dockerfile
+++ b/ds-rshiny/files/docker/Dockerfile
@@ -1,9 +1,12 @@
-FROM ubuntu:16.04
+FROM ubuntu:20.04
 
 ENV DEBIAN_FRONTEND noninteractive
 ENV INITRD No
+ENV LANGUAGE en
+ENV LC_ALL en_US.UTF-8
+ENV LANG en_US.UTF-8
 
-# Dockerize container
+# Dockerize container and setup required OS dependencies
 RUN echo 'force-unsafe-io' > /etc/dpkg/dpkg.cfg.d/02apt-speedup && \
     echo 'DPkg::Post-Invoke {"/bin/rm -f /var/cache/apt/archives/*.deb || true";};' > /etc/apt/apt.conf.d/00no-cache && \
     /usr/bin/dpkg-divert --local --rename --add /usr/bin/ischroot && \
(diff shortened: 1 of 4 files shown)


ds-rshiny/files/docker/Dockerfile
```diff
@@ -35,25 +36,30 @@ RUN apt-get update && \
...
-RUN R -e 'packages = c("RJDBC", "shinythemes", "ggplot2", "reshape2"); \
+RUN R -e 'packages = c("shinythemes", "ggplot2", "reshape2"); \
     install.packages(packages, repos="http://cran.uni-muenster.de/"); \
     package.check <- lapply(packages, FUN = function(p) { if(!require(p, character.only = TRUE)) { stop("package not found!") } })'
 
 # Cleanup apt
-RUN apt-get clean && rm -rf /var/lib/apt/lists/*
+RUN apt-get autoremove --purge && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* /var/cache/apt/archives
```
michaelsauter, line 57: This is a bit pointless actually. Given container images are layers, this does not actually make the image any smaller. I see it's already been like that before though. I'd either remove it, or, if possible, add it to the layer which runs `apt-get install`.

gerardcl: Makes sense! on the other side I see it is more to ensure that later on noone can make use of apt.

Will give it a try and come back! 👍

michaelsauter: Whether one can use `apt` is more a matter of permissions - if I'm not mistaken, even on OpenShift in building you can switch back to root and then do `update` and `install` all over again ;)

felipecruz91 (approved):

ds-rshiny/files/docker/Dockerfile
gerardcl: ok, I did some refactor with success and it really looks much more better! thanks for taking a look @michaelsauter !! 🚀

michaelsauter (approved):

PR #592, opened 2021-05-14
Add recommended Kubernetes labels

### LINKED ISSUE #512: Add recommended Kubernetes labels
https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/ describes a set of recommended labels for each resource:

> You can visualize and manage Kubernetes objects with more tools than kubectl and the dashboard. A common set of labels allows tools to work interoperably, describing objects in a common manner that all tools can understand.
>
> In addition to supporting tooling, the recommended labels describe applications in a way that can be queried.

I think it would make sense if ODS would add those labels to every provisioned quickstarter. I would suggest the values as follows:
```
app.kubernetes.io/name: ${COMPONENT_ID}
app.kubernetes.io/instance: ${COMPONENT_ID}
app.kubernetes.io/version: latest
app.kubernetes.io/component: be/fe/ds
app.kubernetes.io/part-of: ${PROJECT_ID}
app.kubernetes.io/managed-by: oc
```

A couple of notes on this:

* `app.kubernetes.io/instance` should equal `app.kubernetes.io/name` by default I think, but this label can be used to differentiate multiple instances of the component in the same namespace (e.g. using the approach described at https://www.opendevstack.org/ods-documentation/opendevstack/3.x/jenkins-shared-library/component-pipeline.html#_deploying_openshift_resources_from_source_code, or in the future using a different Helm release, see also their recommendations at https://helm.sh/docs/chart_best_practices/labels/)
* `app.kubernetes.io/version` could be updated with the Git commit during each Jenkins run? I think it would be nice to do this, but it conflicts with the "export of live configuration" feature in the orchestration pipeline. 
* `app.kubernetes.io/component`: this could be set depending on the quickstarter, which already has information what "kind of component" it is.
* `app.kubernetes.io/managed-by`: this is a tricky one, as by default we create the resource once but then it is manually managed. The label is used by Helm to mark all resources managed by Helm. We could either not define it initially, or set it to one of `manual`, `oc` or `tailor`. Thoughts?
* How would we deal with existing components? Should we add those labels to each resource during every Jenkins run, or should we add a section in the upgrade guide?
* There is some overlap with the existing `app=${PROJECT_ID}-${COMPONENT_ID}` label, but I would initially not touch that and just have both things side-by-side.

Asking for input @metmajer @clemensutschig @henrjk @gerardcl. Do you have thoughts on this? Do you agree with the proposed values?

jafarre-viewnext:
Adapted the metadata.yml of be-gateway-nginx quickstarter so that the appropriate labels will be automatically set to newly-provisioned components.

Closes #512

Tasks: 
- [X] Adapted metadata.yml of the be-gateway-nginx quickstarter
- [X] Added comments to document the usage of the different metadata entries

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - requests access logging enabled for openshift oauth proxy component (used by ds-rshiny and ds-jupyter-lab) ([#590](https://github.com/opendevstack/ods-quickstarters/issues/590))
+- Add recommended Kubernetes labels ([512](https://github.com/opendevstack/ods-quickstarters/issues/512))
 
 ### Changed
 
--- a/be-gateway-nginx/files/metadata.yml
+++ b/be-gateway-nginx/files/metadata.yml
@@ -11,6 +11,9 @@ supplier: https://openresty.org
 # Version of the software deployed in this component.
 version: 1.19.3
 
+# Version of OpenDevStack
+odsVersion: 4.x
+
 # Type of component. One of ods, ods-service, ods-test or ods-infra.
 type: ods-service
 
--- a/be-gateway-nginx/files/metadata.yml
+++ b/be-gateway-nginx/files/metadata.yml
@@ -1,20 +1,8 @@
 ---
-# Human-friendly name of the software deployed in this component.
 name: nginx
-
-# Description of the component.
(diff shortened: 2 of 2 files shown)


jafarre-viewnext:
As agreed with @metmajer , no modifications will be done to the quickstarters.
Closing this pull request.

PR #613, opened 2021-06-29
remove Pipfile.lock as it is not used and not critical for the environment

tbugfinder:
Drop Pipfile.lock from inf-terraform-aws quickstarter component.
It's neither used, nor core for the qs.

Tasks: 


metmajer:
Thanks @tbugfinder! According to your task list, a documentation update seems to be missing. Hmm?

tbugfinder:
will come back with a single PR.

PR #612, opened 2021-06-29
be-gateway-nginx switch from centos to fedora

gerardcl:
Fixes #611 

Tasks: 
- [x] Updated documentation in `docs/modules/...` directory
- [x] Ran tests in `<quickstarter>/testdata` directory

PR #621, opened 2021-07-07
be-gateway-nginx maintenance upgrade

### LINKED ISSUE #611: be-gateway-nginx switch from centos
We are currently using centos7 base image for the be-gateway-nginx quickstarter. Since centos is not not any longer maintained, and althought the current quickstarter works, we will face obvious issues soon and also some issues might happen already when trying to update the repos or adding other packages for example.

We can just switch to fedora since that is also now what the openresty project has done and is providing as base image.

Keep current centos7 or switch to a full different distro.

**Additional context**
N/A

gerardcl:
Fixes #611
Relates to #612

Tasks: 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 - ds-rshiny cleanup cloudera dependency ([#540](https://github.com/opendevstack/ods-quickstarters/pull/540))
 - ds-rshiny quickstarter goes to broken repository ([#605](https://github.com/opendevstack/ods-quickstarters/issues/605))
+- be-gateway-nginx switch from Alpine to Fedora - backport from master/4.x ([#611](https://github.com/opendevstack/ods-quickstarters/issues/611))
 
 ## [3.0] - 2020-08-11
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 
 - ds-rshiny cleanup cloudera dependency ([#540](https://github.com/opendevstack/ods-quickstarters/pull/540))
 - ds-rshiny quickstarter goes to broken repository ([#605](https://github.com/opendevstack/ods-quickstarters/issues/605))
+- be-gateway-nginx switch from CentOS to Fedora ([#611](https://github.com/opendevstack/ods-quickstarters/issues/611))
```
michaelsauter, line 14: switch is from alpine, right? Is the alpine image not maintained anymore?

gerardcl: Good catch! just backported all to 3.x, in 3.x we were still using alpine and was fully unmaintained, we migrated to centos in the past and that was not backported to 3.x. The reason to move to centos in the past (and now fedora) was mainly due to better OS dependencies support.

PR #622, opened 2021-07-07
fix dead sbt rpm bintray repo

### LINKED ISSUE #619: Scala agent build fail, it is forbidden to download sbt-1.3.13.rpm
**Describe the bug**
Building the scala agent fails when it tries to download sbt rpms.

If you try to download the rpm as it is done in the quickstarter 'curl -O -L http://dl.bintray.com/sbt/rpm/sbt-$SBT_VERSION.rpm' it fails as it is forbidden to download the file.

**To Reproduce**
1. Rebuild the scala agent.
2. Build will fail, and if you check the downloaded file it will contain "Forbiden"

**Expected behavior**
The image is built

 - OpenShift: [ALL]
 - OpenDevStack [ALL]

```
Step 8/19 : RUN curl -O -L http://dl.bintray.com/sbt/rpm/sbt-$SBT_VERSION.rpm  &&     yum -y install sbt-$SBT_VERSION.rpm &&     yum clean all &&     rm -rf /var/cache/yum
 ---> Running in 4120d8ef5e99
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0100    10  100    10    0     0    151      0 --:--:-- --:--:-- --:--:--   153
Loaded plugins: fastestmirror, ovl
Cannot open: sbt-1.3.13.rpm. Skipping.
```

oalyman:
This PR fixes the issue that scala agents could not be build anymore due to the bintray sbt rpm repo not available anymore (https://jfrog.com/center-sunset/).

Furthermore it activates the build of the scala agents during execution of the CI via the github actions.
This build was disabled (thus we missed the problem with the dead rpm repo) due to the way the scala agent was build previously.
It included a test that verified, that sbt could be downloaded via nexus during the test image build.
Now a more lax test is executed, that allows us to verify that sbt was installed but excludes the nexus test.
No other agent for the build tools of ods verifies this by the way so I think it is the right tradeoff.

Closes #619 
Fixes #619

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -150,20 +150,36 @@ jobs:
         run: |
           docker build --tag agent-terraform-test-ubi8 --file Dockerfile.ubi8 .
 
+  jenkins-agent-scala-rhel7:
+    name: Jenkins agent Scala (RHEL7)
+    runs-on: ubuntu-18.04
+    steps:
+      -
+        name: Checkout repository
+        uses: actions/checkout@v2.0.0
+      -
+        name: Build docker image
+        working-directory: common/jenkins-agents/scala/docker
+        run: |
+          docker build --tag agent-scala-test-rhel7 --file Dockerfile.rhel7 \
+             --build-arg nexusUrl=https://nexus.example.com \
+             --build-arg nexusUsername=developer \
+             --build-arg nexusPassword=s3cr3t \
+             .
 
-# jenkins-agent-scala:
-#   name: Jenkins agent Scala
-#   runs-on: ubuntu-18.04
-#   steps:
-#     -
-#       name: Checkout repository
-#       uses: actions/checkout@v2.0.0
-#     -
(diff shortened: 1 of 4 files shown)


clemensutschig (approved):

clemensutschig:
@oalyman cool. The reason i did this was to have the base dependencies loaded during build of the image... and it was a great test zu make Sure nexus really Works..

oalyman:
@clemensutschig I know and I still think it's a good idea. An option would be to be more flexible how to set the repo proxy configuration so that when building in github we could use the sbt default repos directly and when configuring nexus settings using those.
Maybe by using a script comparable to how the proxy is set...

michaelsauter (approved):

clemensutschig:
@oalyman  - why can't we have a switch, if nexus setup the nexus pieces in the image - and if not - then not .. (and still run sbt build)

PR #650, opened 2021-09-14
Bugfix/update external dep to3x

s2oBCN:
Update external dependencies


metmajer:
@s2oBCN we will not release 3.x anymore. Backports need to happen into 4.x.

s2oBCN:
Then we should remove the CI. Is this ok?

Issue #651, opened 2021-10-11
Use Java 17 (LTS) in maven jenkins-agent and spring boot quickstarter


renedupont:
Java 17 has been released a couple of months ago. Since it is an LTS version, it might make sense to use it. Currently we are at Java 11. 
Once we on it, we could also update gradle in the spring boot quickstarter, we are 3 major versions behind.

Any reason what would speak against it? Too early?

renedupont:
@metmajer yes makes sense, that's what I "feared" 😄 

@stitakis I was also refering to the (maven) agent which does
`FROM opendevstackorg/ods-jenkins-agent-base-ubi8:latest`
and then adds java 11 (https://github.com/opendevstack/ods-quickstarters/blob/master/common/jenkins-agents/maven/docker/Dockerfile.ubi8)
without the agent being on 17 we can't build them on it.

metmajer:
@stitakis if you do not wait, it means that the owners of this quickstarter will have more work sorting out security incidents themselves vs. being supported by a vendor.

renedupont:
@stitakis just to clarify, you meant regarding eclipse-temurin not the new official open jdk source, but rather new **adopt** open jdk, right?

metmajer:
@stitakis what about you make the Quickstarter with jdk 17 available on a feature branch that you release into the BIX cluster and leave both master and 4.x intact? Just to clarify, with security scanning being activated on all new clusters, we want to reduce the likelihood for vulnerabilities. The best way to keep the house clean seems to be that all quickstarters are based on a common, secure foundation. Not all quickstarteres are there yet, but we need to get there very soon.

stitakis:
@metmajer ok, I see... it looks to early to upgrade to jdk 17 based on an ubi image

renedupont:
> @renedupont but, does it need to be on the same jdk version? btw... I can't remember, why maven is needed to be installed in the agent?

Correct me if I'm wrong, but if you write your app in Java 17 but try to build it with an agent that only has Java 11 installed, it will fail.

stitakis:
@kuebler @renedupont I have merged this to master. I'll port this feature to 4.x

PR #675, opened 2021-11-23
Feature/gh 651 upgrade java to version 17

stitakis:
implements #651


stitakis:
@renedupont this a draft PR, it still needs to be tested. I'm not sure if it complete, could you please have a quick look?

common/jenkins-agents/maven/docker/Dockerfile.centos7
```diff
@@ -23,9 +23,9 @@ ENV BASH_ENV=/usr/local/bin/scl_enable \
     GRADLE_USER_HOME=/home/jenkins/.gradle
 
 # Install Java
-RUN yum install -y java-11-openjdk-devel && \
-    yum clean all -y && \
-    exactVersion=$(ls -lah /usr/lib/jvm | grep "java-11-openjdk-11.*\.x86_64" | awk '{print $NF}' | head -1) && \
+RUN yum install -y epel-release && yum repolist && yum install -y java-latest-openjdk-devel && \
```
stitakis, line 26: This install the EPEL repository (Extra Pacakges for Enterprise Linux, see https://www.cyberciti.biz/faq/installing-rhel-epel-repo-on-centos-redhat-7-x/), otherwise the package `java-latest-openjdk-devel` is not found. Opposed to centos 8, the java package name in centos 7 called `latest`.

common/jenkins-agents/maven/docker/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Tue Jul 03 16:54:12 CEST 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3-bin.zip
```
stitakis, line 3: Time to upgrade to latest gradle version. We could also upgrade maven

be-java-springboot/files/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM adoptopenjdk/openjdk11:alpine-jre
+FROM registry.access.redhat.com/ubi8/openjdk-17
```
stitakis, line 1: switching to redhat ubi8 image

metmajer: Like it, thank you

michaelsauter:
Cool!

If I get this right, this moves JDK from 11 -> 17 and Gradle from 6 -> 7. I wonder what effect this has on the user? I would love to have this detailed a bit more, and recorded somewhere user-facing. In the past we collected this information in partials (e.g. https://github.com/opendevstack/ods-quickstarters/blob/master/docs/modules/quickstarters/partials/notes-4x.adoc).

Also - we should think about when to roll this out ... so far the policy was "breaking change = major version" and since quickstarters are not versioned independently from ODS core, this might mean ODS 5 as a target release if we consider JDK 17 to be breaking. TBH I am not entirely sure what the backwards compatibility promise of JDK 17 is exactly. Do you have more info on this?

be-java-springboot/files/metadata.yml
```diff
@@ -1,6 +1,6 @@
 ---
 name: OpenJDK
-description: "OpenJDK is a free and open-source implementation of the Java Platform, Standard Edition. Technologies: Spring Boot 2.1, OpenJDK 11"
-supplier: https://adoptopenjdk.net
+description: "OpenJDK is a free and open-source implementation of the Java Platform, Standard Edition. Technologies: Spring Boot 2.1, OpenJDK 17"
```
stitakis, line 3: @michaelsauter I have updated this file, but I'm sure why we have it. For which purpose this file is used?

michaelsauter: @metmajer Is there documentation on this?

I believe this to be used in document generation as a description of the component.

be-java-springboot/files/docker/Dockerfile
michaelsauter: Note https://www.redhat.com/en/blog/transitioning-red-hat-container-registry. TBH I am not really sure what we should do about it. We can use `registry.redhat.io` instead, but that would break local builds. Another approach is to replace the `FROM` on the fly (via the `BuildConfig`) but that might need some ground work first.

michaelsauter:
@stitakis Yes that would certainly be an option, and one that is very smooth! However personally I am not in favour of this. Here is why.

My view is derived from the following considerations, which are also mentioned in https://github.com/opendevstack/ods-pipeline/issues/211. Note that ODS Pipeline has a huge advantage here because we version it independently from ODS core.

* I'd rather offer few choices that work well, compared to many choices that are half-baked. Unfortunately, I think the quickstarters in this repo already vary greatly in quality. Therefore my vote basically goes by default to "fewer options".
* Evolving versions of programming languages always cause effort somewhere, the question is just who pays the price. I think the ODS open-source project should not pay the maintenance effort and instead strive to keep up-to-date with new version as much as possible/reasonable (e.g. tracking LTS releases).
* Let's assume for now we run with this approach. That gives the next team in the chain (the platform owners / the people that installed ODS in the cluster) the choice to keep the ODS 4 version (and therefore ODS 4 quickstarters) around after an update to ODS 5, or to even maintain a local "ODS 5 version" of the JDK 11 quickstarter. Whenever that team decides to remove ODS 4, it is up to the next team in the chain (a project team) to deal with this situation. They can either update their application, or maintain a local "ODS 4 (or 5) version" of the JDK quickstarter.
* Apart from this, my personal view is that backwards compatibility should factor into the choice of language / environment in project teams. If I want to avoid upgrade effort, then I should pick a language with a matching profile (e.g. Go with its [compatibility promise](https://go.dev/doc/go1compat), and the JDK is also quite stable I think). If I do not pick such a language or am unable to for other reasons, I need to make this transparent as a team and need to invest in upgrade activities continuously.

Again, this is my personal view only. @metmajer I think it would be helpful to define our approach somewhere :)

stitakis:
@michaelsauter this scenario comes to my mind. Let say we upgrade in ODS the java to version 17 in the maven jenkins agent and ODS. How would a product team upgrade a product (assumming from my experience that effort takes some weeks, and some release to adapt) which in production to a new java version if compiling with java jdk 17 is not an option. I see the need to provide an upgrade path for this scenario. How would this work if we do not provide 2 maven agents, one with java 11 and one with java 17?

stitakis:
> ... How would this work if we do not provide 2 maven agents, one with java 11 and one with java 17?

The standard way to upgrade projects from one ODS version to new one is by changing the jenkins agent defined in the project Jenkinsfile. Thanks @michaelsauter for clarifying this.

This approach will work for most of the projects. However it will not work, if a project have to stay in java 11 but still want to benefit from the new ODS version.

To solve this limitation, I have added 2 shell scripts that switches the java version to this PR.
These scripts can be used in the project Jenkinsfile. E.g. to switch to java 17 in the project Jenkinsfile the script should call `use-j17.sh` before calling gradlew. Here an example: 
```
stage('Build and Unit Test') {
    withEnv(["TAGVERSION=${context.tagversion}", "NEXUS_HOST=${context.nexusHost}", "NEXUS_USERNAME=${context.nexusUsername}", "NEXUS_PASSWORD=${context.nexusPassword}", "JAVA_OPTS=${javaOpts}","GRADLE_TEST_OPTS=${gradleTestOpts}","ENVIRONMENT=${springBootEnv}"]) {
      def status = sh(script: "use-j17.sh && ./gradlew clean build --stacktrace --no-daemon", returnStatus: true)
      if (status != 0) {
        error "Build failed!"
      }
    }
  }

```
I have also update the `be-java-springboot` quickstarter to switch to java 17 before calling maven. New spring boot quickstarters will use java 17 by default.

I think this approach addresses some goals. On one side, this solves the overhead of having to maintain 2 maven agents that only differ in the java version installed. Or the other side, it allow java projects to migrate first to new ODS version and afterwards to migrate to the new java version.

@metmajer what do you think?

metmajer:
@stitakis @michaelsauter I will need some more days to dig into this. The question I have is: how can we give people a choice? 

- If new projects need to use the latest version of a particular technology, that's mostly fine, if the solution is self-developed. What if a solution is not self-developed? Vendors might request a particular version version of a runtime to guarantee support.
- If existing projects want/have to migrate to the latest version of OpenDevStack, including a migration of their components, breaking changes in the runtime will likely break their solution.

I am not done with my thinking yet, but I wonder if we should offer multiple Quickstarter versions around the same technology? Since Quickstarters depend on the Jenkins shared library, we always have the potential of a breaking change in there, e.g. a component Stage being changed or deleted in the shared library which affects an existing Quickstarter that offers JDK11 and was built for ODS 3.

renedupont:
> * I'd rather offer few choices that work well, compared to many choices that are half-baked. Unfortunately, I think the quickstarters in this repo already vary greatly in quality. Therefore my vote basically goes by default to "fewer options".
> * Evolving versions of programming languages always cause effort somewhere, the question is just who pays the price. I think the ODS open-source project should not pay the maintenance effort and instead strive to keep up-to-date with new version as much as possible/reasonable (e.g. tracking LTS releases).

In at least 3 of the last 4 projects I have been part of we created a custom jenkins agent and I've seen others doing it aswell. I think the reason for that is, that it is just a low chance that a given agent already includes every tool and dependency etc. you need in your specific case. Because of that, providing "fewer options" might makes sense, because less has to be maintained and the likelihood of requiring a custom agent is high. That might just be my limited observation though.

So I'm wondering if there is any downside of providing an easier way to create and use custom agents? I mean according to this: https://github.com/opendevstack/ods-quickstarters/pull/586/files it is not super hard to do currently, the question for me is, if this is less desired from a platform point of view. The responsibility of maintaining the agents is then more on the developer's side. Is this a security risk because they might not update them as frequently?

> * Apart from this, my personal view is that backwards compatibility should factor into the choice of language / environment in project teams. If I want to avoid upgrade effort, then I should pick a language with a matching profile (e.g. Go with its [compatibility promise](https://go.dev/doc/go1compat), and the JDK is also quite stable I think). If I do not pick such a language or am unable to for other reasons, I need to make this transparent as a team and need to invest in upgrade activities continuously.

Yes, JDK versions should be compatible within a major version as Go promises aswell, its just that since a few years they started to release new java major versions more frequently compared to earlier times 😆. There is btw. this neat page that lists and compares every Java API change by versions: https://javaalmanac.io/

In general I'd find it too risky to just roll out a new major Java version in the agents and expecting consuming teams to change something until everything works for them again like before (and just be it updating the JDK in their local env). So in case we do not want people to create custom agents, I'd rather go with Sebastian's suggestion to sort of create a new agent for the new LTS Java version and deprecate the agent with the old Java LTS version and remove it with upcoming ODS releases, even if that means having two mostly identical agents at a time.

be-java-springboot/Jenkinsfile.template
```diff
@@ -26,7 +26,7 @@ def stageBuild(def context) {
   }
   stage('Build and Unit Test') {
     withEnv(["TAGVERSION=${context.tagversion}", "NEXUS_HOST=${context.nexusHost}", "NEXUS_USERNAME=${context.nexusUsername}", "NEXUS_PASSWORD=${context.nexusPassword}", "JAVA_OPTS=${javaOpts}","GRADLE_TEST_OPTS=${gradleTestOpts}","ENVIRONMENT=${springBootEnv}"]) {
-      def status = sh(script: "./gradlew clean build --stacktrace --no-daemon", returnStatus: true)
+      def status = sh(script: "use-j17.sh && ./gradlew clean build --stacktrace --no-daemon", returnStatus: true)
```
renedupont, line 29: Interesting idea... I mean you could even check the jar artifact for the target jdk it was built for and decide depending on that automatically, but for me these solutions are a bit hacky and also make traceability more complex.

stitakis: @renedupont I can't follow. The jar artifact does not exists at this point. It will be created by calling gradle. Can you please elaborate a bit more?

btw... have a quick look at the script `use-j17.sh`. It is just 4 lines. It is effective and short and same instructions used in dockerfile to set the java  version.

be-java-springboot/files/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.5.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3-bin.zip
```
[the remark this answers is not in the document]
stitakis: @renedupont this gradle file will be used by a new spring project after the quickstarter is provisioned. We need to test a new java spring project based on this quickstarter with this gradle configuration. I'll do a test and come back on this. Or do you any other scenario of a potential breaking change?

stitakis:
> If new projects need to use the latest version of a particular technology, that's mostly fine, if the solution is self-developed. What if a solution is not self-developed? Vendors might request a particular version version of a runtime to guarantee support.

The maven agent is used to compile java application. If we have a vendor that provides code that need to be compiled with a special jdk version and special jenkins agent could be build. But I see as a corner-case scenario that we don't aim to cover with the maven agent.

>If existing projects want/have to migrate to the latest version of OpenDevStack, including a migration of their components, breaking changes in the runtime will likely break their solution.

This is specially the case I'm covering with this PR. It provides a smooth-path for this. It supports jdk 11 as default and the latest long-term-support version (LTS) java 17 as optional. If you need to compile with java 17 you just add "use-j17.sh && ./gradlew` before calling gradle.

> I am not done with my thinking yet, but I wonder if we should offer multiple Quickstarter versions around the same technology? Since Quickstarters depend on the Jenkins shared library, we always have the potential of a breaking change in there, e.g. a component Stage being changed or deleted in the shared library which affects an existing Quickstarter that offers JDK11 and was built for ODS 3.

Good point. At moment the current jenkins agent runs only with `java 11`. For this reason the default java version of the agent is still java 11. This PR enable users to change the java version to java 17 as shown above to be able to compile with latest java. The jenkins-shared-lib is not impacted at this moment because it was previously loaded and runs in a different process.

be-java-springboot/files/gradle/wrapper/gradle-wrapper.properties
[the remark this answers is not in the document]
renedupont: Oh you're right... this only relates to new provisioned components!

be-java-springboot/Jenkinsfile.template
renedupont: Oh yes, somehow I was only thinking about the jdk version of the docker container that runs the app later on... so nevermind my comment. 

But some other things just came up my mind: 
- If I understand it right, we link the java + javac version depending on which script I use and it will be build for that version due to that. But what if someone specifically uses in his `build.gradle` something like:

```
apply plugin: 'java'

compileJava {
    sourceCompatibility = 17   // not sure about right syntax :)
    targetCompatibility = 17
}
```
but picks use-j11.sh ... do we just treat this as a user misconfiguration?

- Similar to the previous one, since gradle 6.7 you can do
```
java {
    toolchain {
        languageVersion.set(JavaLanguageVersion.of(11))
    }
}
```
which does according to the [doc](https://docs.gradle.org/current/userguide/toolchains.html#sec:consuming):

> 1. Setup all compile, test and javadoc tasks to use the defined toolchain which may be different than the one Gradle itself uses
> 2. Gradle detects locally installed JVMs
> 3. Gradle chooses a JRE/JDK matching the requirements of the build (in this case a JVM supporting Java 11)
> 4. If no matching JVM is found, it will automatically download a matching JDK from AdoptOpenJDK

which is also another possibility to determine the target java version next to the use-jxx.sh scripts. Do you see that as an issue? 

Btw. What I like about this toolchain thing is that you could potentially do a build with every java version you like to independently of the jenkins-agents version (which would solve the general issue we have in this thread)... but on the downside this would download the adoptopenjdk with every pipeline run in case it differs from the one installed on the agent. I guess the downside is worse here, what do you think? Also this does not help if people use maven over gradle.

stitakis: I had no problem in offering and maintaining two agents in ODS. I'm proposing a solution for one agent because Michael raise the maintenance cost as issue. Anyway at the moment the `use-j17.sh` provide a simple straightforward solution which I believe is easy to maintain and is not complex at all.

 > but picks use-j11.sh ... do we just treat this as a user misconfiguration? The issue is that we now have more than one place to determine the java version it should be built with.

gradle will fail in this case. This PR upgrades the spring boot quickstarter to use `java 17` and adds `use-j17.sh` in the Jenkinsfile to switch the java version. The default will be java 11, so that backwards compatibility is guaranteed for project upgrading to ODS 5.x

> 2. Similar to the previous one, since gradle 6.7 you can do...

I don't this as an issue at the moment. The problem of downloading the jdk on the fly is the performance penalty on the build duration.

@renedupont do you see any blocker, or are you fine with this PR? I'd like to convert from draft to a real PR soon.

renedupont (approved):

metmajer:
@stitakis I feel this will impact existing users, qualification and migration guidelines. Final solution needs to be enterprise ready, too. Let’s discuss the end user experience based on my previous questions soon — and do not merge yet.

be-java-springboot/files/docker/Dockerfile
stitakis: this doesn't sound good. I think it would be better to stick with adoptopenjdk until the red hat registry issue is being clarified.
I'll change the image to `eclipse-temurin:17-jdk-alpine` which is and uses the java adoptium distribution from Adoptium, which is the official and equivalent successor of adoptopenjdk (see: https://adoptium.net/)

stitakis:
@metmajer yes, let us discuss. I'll schedule a meeting with you to go through the open points.

As preparation for our discussion I have summarized the changes provided by this branch below.

These draft PR add following changes to

1. jenkins maven agent:
- installs additionally java 17 to the jenkins maven agent. Java 11 still the default version at runtime (backward compatibility guaranteed)
- adds script `use-j17.sh` to allow users to opt to compile java projects with java 17. (backward compatibility guaranteed) 

2. java-springboot-quickstarter
- upgrades spring boot version to current lastest 6.2.1
- add to Jenkinsfile.template call to `use-j17.sh` before calling gradle wrapper, so that new projects based on this quickstarter can use java 17.
- upgrades docker image to java 17

Apart from missing documentation I think this PR is enterprise ready because:
- in term of user experience this change is an additive change and therefore fully backward compatible.
- in term of validation I don't see any extra risks provided by java 17 or latest spring boot. Only the documentation should be updated to reflect the additional java version and the update of spring boot version.

btw... I have few questions related to discuss:
> If new projects need to use the latest version of a particular technology, that's mostly fine, if the solution is self-developed. What if a solution is not self-developed? Vendors might request a particular version version of a runtime to guarantee support.

As I see this, the case of a solution for not self-developed is not related to the jenkins maven agent. The agent is used to compile java software project. Vendor solutions will likely be integrated and deployed using the docker quickstarter by importing there the vendor image, isn't it? Or do I missing something?

> If existing projects want/have to migrate to the latest version of OpenDevStack, including a migration of their components, breaking changes in the runtime will likely break their solution.

As mentioned above, this PR is fully backward compatible and do not introduce any breaking changes at all, neither to the jenkins maven agent runtime nor to project migrating to this agent.

metmajer:
Please document how Java 17 can be enabled in the official ODS documentation for ODS 5. Otherwise good to go!

metmajer:
@stitakis @renedupont notes from our discussion today:

**Goals for Quickstarters and Jenkins Agents in OpenDevStack**
- Easily bring applications using old versions of ODS to its latest version
- Provide a choice around the build/runtime versions (e.g. JDK 11 + 17 + ...)

**Challenges**
- Hard dependency on SonarQube. The [Jenkins base Agent contains a specific version of the SonarQube Scanner CLI](https://github.com/opendevstack/ods-core/blob/master/jenkins/agent-base/Dockerfile.ubi8#L29-L35) which needs to be compatible with the SonarQube service that comes with the latest version of ODS. If versions are not compatible, the SQ integration will break.

**Solution?**
- Build dependencies and runtime dependencies are always contained in Jenkins Agents
- Every new agent includes the build/runtime versions of the previous agents + a version of the SonarQube Scanner CLI that is compatible with the SonarQube service's particular version. Reason: allow application teams to use the latest version of ODS and not miss their favorite version of a particular Quickstarter technology.
- With each new ODS version, existing Jenkins Agents need to be rebuilt. Reason: ensure continued service with SonarQube and avoid incompatible versions between Scanner CLI and service.

stitakis:
@michaelsauter @renedupont @jorge-romero @jafarre-viewnext I have created this new issue #694 to address the feature request defined @metmajer in previous comment.

be-java-springboot/files/metadata.yml
metmajer: @stitakis @michaelsauter yes, this information will be used in LeVA documentation. If this says OpenJDK 17, it must correspond with what's actually used.

be-java-springboot/files/docker/Dockerfile
stitakis: @michaelsauter we need your feedback.... can you please comment on this?

stitakis: @michaelsauter thanks for the feedback. No I haven't read about an activation date of oauth from red hat. Since this discussion is still open and not fully discussed I think it is fine to leave at it is. We can change the docker image later on to a unified one.

stitakis:
@renedupont @kuebler @metmajer @michaelsauter  this is ready for review. I have extended the documentation including a migration guide for projects updating to java 17. It would be awesome if we can close this soon.

kuebler (dismissed):
LGTM, thanks for putting in all the effort, @stitakis! 🚀

stitakis:
@kuebler I have merged master into this branch. Can you please review again?

kuebler (approved):
LGTM! 🚀

stitakis:
@kuebler @renedupont I have introduced some bug fixes after testing this quickstarter. Can you please review again?

kuebler (approved):
LGTM 🚀  IMO, check failures for the terraform agent look unrelated to this changes

PR #684, opened 2021-12-03
Bug/missing dir 679

### LINKED ISSUE #679: AWS Quickstarter: reports/installed directory missing
When deploying the QS with the prov app the reports/install directory will not be deployed though it is part of the quickstarters files directory.

reports/install is used to keep some temporary files for the document creation.

it seems we run into some sort of chicken egg issue because though there is a .gitkeep in this directory reports/install itself is mentioned in the .gitignore file as well.

**Affected version :**
 - aws quickstarter [4.x]

nichtraunzer:
Add proper config to gitignore to keep empty log dir

Fixes #679

Tasks: 
- [X] Ran tests in an EDP in a Box.


tbugfinder (approved):
LGTM

PR #699, opened 2022-01-11
updates java to version 17 and gradle to version 7.3 port to 4.x

### LINKED ISSUE #651: Use Java 17 (LTS) in maven jenkins-agent and spring boot quickstarter

stitakis:
fixes #651 port 4.x


kuebler (approved):
LGTM! 🚀

stitakis:
@kuebler fyi, in my last commit I've added the jacoco bufgix (refs #700 and #701) otherwise the end to test (ODS AMI build) fails.

renedupont:
looks good so far,  just had a couple of notes

- Is there a specific reason why you have different gradle versions for the springboot qs and for the agent (and both are not the latest btw)?
- Spring Boot is meanwhile on 2.6.2 😆

renedupont (approved):

stitakis:
@renedupont no special reason for the different versions of gradle... I'll create another PR to use the same version everywhere and also upgrade spring boot to the latest one.

stitakis:
@renedupont feel free to submit a PR if you like to harmonize the gradle version and upgrade to latest spring boot

PR #726, opened 2022-02-18
fix agents from jenkins agent base repo changes

gerardcl:
This PR contains fixes to the agents affected by the change in https://github.com/opendevstack/ods-core/pull/1106

Fixes  #722

Tasks: 
- [x] Updated documentation in `docs/modules/...` directory
- [x] Ran tests in `<quickstarter>/testdata` directory

PR #727, opened 2022-02-18
fix scala,nodejs12 and maven agents from jenkins agent base repo changes

gerardcl:
port from https://github.com/opendevstack/ods-quickstarters/pull/726

Tasks: 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,9 @@
 
 ### Fixed
 - inf-terraform-agent: fix pip update and epel installation
-
+- Mavent agent updated from Jenkins base image changes ([#722](https://github.com/opendevstack/ods-quickstarters/issues/722))
+- NodeJS12 agent updated from Jenkins base image changes ([#720](https://github.com/opendevstack/ods-quickstarters/issues/720))
+- Scala agent updated from Jenkins base image changes ([#721](https://github.com/opendevstack/ods-quickstarters/issues/721))
 
 ## [4.0] - 2021-05-11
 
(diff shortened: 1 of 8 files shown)


tbugfinder (commented):
I'd drop CentOS8 and move on towards rockylinux repos or AlmaLinux repos.
Additionally, the base agent has CodeReady Builder repo enabled which depends on a valid subscription.

metmajer:
@gerardcl @tbugfinder we'll need to switch to images which are regularly scanned and fixed by a vendor, e.g. UBI8 since otherwise security scanners will constantly alarm about vulnerable images.

tbugfinder:
> @gerardcl @tbugfinder we'll need to switch to images which are regularly scanned and fixed by a vendor, e.g. UBI since otherwise security scanners will constantly alarm about vulnerable images.

ubi8 repos do not provide all packages in use, hence additional repos are needed. RHEL clone projects build packages after upstream releases, timely.

metmajer:
Thanks @tbugfinder for the hint!

gerardcl:
hi @tbugfinder IIRC once there is a yum interaction if the tool detects you have a subscription then auto manages repos. In any case, the codeready repo is already available via the `ubi.repo` (see https://github.com/opendevstack/ods-core/blob/4.x/jenkins/agent-base/yum.repos.d/ubi.repo).

My suggestion would be, in case you need the centos8 stream repos in the terraform agent, is on doing something similar done as in the maven or nodejs12 agents; since they both depend on packages still not in ubi.repo, by just COPY the centos8.repo in then all good. See https://github.com/opendevstack/ods-quickstarters/pull/727/files#diff-9dc8b4be0b5c186d69d81fc38a2f3b7b0bb574002158d9d103a31993380e571eR32 and https://github.com/opendevstack/ods-quickstarters/pull/727/files#diff-9950561929f5cd61eab8d1fa8bb6c9658ef9cf9de9fdf1f75d4d561a67dd0ca5R41

This is giving us much more fast adaptability to changes in future in the current main branches (4.x and master).

tbugfinder (approved):
CentOS8-Stream isn't my favorite, though :-D

PR #781, opened 2022-04-04
fix and raise spring boot version for initializr

### LINKED ISSUE #779: Update Spring boot qs due to CVE
Hello,

due to https://spring.io/blog/2022/03/31/spring-framework-rce-early-announcement I propose to update the spring boot version to 2.6.6. 
As far as I understood we most likely won't have issues with it because we usually don't package as .war files which seems to be a requirement to be impacted according to https://spring.io/blog/2022/03/31/spring-framework-rce-early-announcement#am-i-impacted but I guess it is still worth upgrading just in case.

renedupont:
Fixes #779 

The current version of spring boot (2.6.1.RELEASE) that we provide to the spring initializr is not valid because that .RELEASE suffix was removed many versions ago from the version name. My guess is that it takes latest instead as default. This PR will fix this and set it to the latest spring boot version.


renedupont:
The current version of spring boot (2.6.1.RELEASE) that we provide to the spring initializr is not valid because that .RELEASE suffix was removed many versions ago from the version name. My guess is that it takes latest instead as default. This PR will fix this and set it to the latest spring boot version.

Issue #506, opened 2020-11-30
Cleanup Jenkins agent image labels


michaelsauter:
We use the labels as specified by RedHat. I think they are only useful for their build service and make no sense for us at all. I think we should just remove them.

While we are at it, we could remove the `maintainer` label too.

Thoughts @faust2199 @gerardcl @henrjk @oalyman?

gerardcl:
Agree! cleanup is always welcome 👍 I think the `maintainer` removal has already been done at some places indeed..

henrjk:
@michaelsauter: Do you mean for example the label defined at https://github.com/opendevstack/ods-quickstarters/blob/883607ea1f2e3fc27e24d36d450b313cc4554f8f/common/jenkins-agents/golang/ocp-config/bc.yml#L25

michaelsauter:
Sorry for the confusion, I meant the image labels in the `Dockerfile`, like https://github.com/opendevstack/ods-quickstarters/blob/master/common/jenkins-agents/maven/docker/Dockerfile.ubi8#L3-L10.

henrjk:
Thanks for the clarification. I am for removing Labels only useful to the RedHat builder. 

To me it appeared that `architecture` and `name` might be useful to keep. 

However I looked what standard labels might be defined and ended up at https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys
I assumed that `LABEL` statement would become an annotation as per the OCI image format.

As architecture is not mentioned anywhere in the annotations it may belong more into the realm of the runtime-spec (https://github.com/opencontainers/runtime-spec).

Any thoughts about platform in particular?

michaelsauter:
Hmm, no particular thoughts around platform.

I think the way I stated it originally was probably too drastic. We can keep `architecture` and `name`. I also found https://bugzilla.redhat.com/show_bug.cgi?id=1334605, which unfortunately leads to a 404, but the `io.k8s` labels could be useful too.

Maybe we start by unifying the labels and remove all `com.redhat` ones?

PR #795, opened 2022-04-25
Add build agent for Node.js 18 (master)

### LINKED ISSUE #794: Provide build agent for Node.js 18
The new LTS version of Node.js - version 18 -  is available now. See also https://nodejs.org/en/about/releases/

The Node.js-specific Jenkins build agents are published per Node.js version. The latest is currently for Node.js version 16 (https://github.com/opendevstack/ods-quickstarters/tree/master/common/jenkins-agents/nodejs16).

For providing a new agent for version 18 I guess the following steps should be fine:
* providing a new agent in the `./common/jenkins-agents` folder similar to the one for Node.js 16
* update build scripts
* update documentation

A very similar update happened recently for the Node.js 16 agent: https://github.com/opendevstack/ods-quickstarters/pull/764

cschweikert:
* add jenkins-agent for Node.js 18
* update chrome to 100.0.4896.127 (latest)
* update documentation

Fixes #794

Tasks: 


common/jenkins-agents/nodejs18/docker/Dockerfile.ubi8
```diff
@@ -0,0 +1,62 @@
+FROM opendevstackorg/ods-jenkins-agent-base-ubi8:latest
+
+# Labels consumed by Red Hat build service
+LABEL com.redhat.component="jenkins-agent-nodejs-18-rhel7-container" \
```
tbugfinder, line 4: isn't it ubi8 instead of rhel7 ?

common/jenkins-agents/nodejs18/ocp-config/bc.yml
```diff
@@ -0,0 +1,66 @@
...
+  required: true
+- name: REPO_BASE
+  required: true
+- name: ODS_IMAGE_TAG
+  required: true
+  value: latest
+- name: ODS_GIT_REF
+  required: true
+- name: JENKINS_AGENT_DOCKERFILE_PATH
+  value: Dockerfile.centos7
```
tbugfinder, line 21: should we default to ubi8 ?

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -129,6 +129,38 @@ jobs:
               --build-arg nexusAuth=developer:s3cr3t \
               .
 
+  jenkins-agent-nodejs18-centos7:
+    name: Jenkins agent NodeJS 18 (CENTOS7)
+    runs-on: ubuntu-18.04
```
tbugfinder, line 134: would 20.04 make sense?

cschweikert: It is `ubuntu-18.04` in all places within this file and I don't know the implications on making the proposed change. Unless the proposed change has a clear benefit or is even a prerequisite for the new build agent, I'd propose to discuss this version change on another issue.

common/jenkins-agents/nodejs18/ocp-config/bc.yml
cschweikert: I'm wondering why we have those two flavors of the Dockerfile in the first place.

@michaelsauter @braisvq1996 Do we still need both?

common/jenkins-agents/nodejs18/docker/Dockerfile.ubi8
cschweikert: Now as I look at this... the labels section definitely needs improvement. Regarding this topic there seems to be #506 for dealing with this on all agents. I would not fix it with this PR.

common/jenkins-agents/nodejs18/ocp-config/bc.yml
braisvq1996: We have both because centos7 was used for Openshift 3 installation and ubi8 for Openshift 4 -> [Check config](https://github.com/opendevstack/ods-core/blob/master/configuration-sample/ods-core.env.sample#L224-L227)
Better maintain both until we decide to drop Openshift 3 completely.

common/jenkins-agents/nodejs18/docker/Dockerfile.ubi8
tbugfinder: I'd prefer to fix it now.

.github/workflows/continuous-integration-workflow.yml
tbugfinder: My thought was, that this nodejs18 agent might live longer than EOL of ubuntu 18.04.

braisvq1996: @tbugfinder We could maintain Ubuntu 18.04 for now and take the upgrade as a separate topic/issue

braisvq1996:
Hi @cschweikert @tbugfinder -> [nodesource](https://github.com/nodesource/distributions#enterprise-linux-based-distributions)
It seems nodejs18 is not working with Centos 7 and Rhel7

cschweikert:
@braisvq1996 good point. Does anything speak against moving on to Centos/RHEL 8? Like already mentioned in one of the discussions above I have no clue about the implications of this switch (e.g. just for the nodejs18 thingy). If it is fine, I can also try that out for now.

braisvq1996:
Switching to Centos8 and/or Rhel8 is a big change.
Right now if we do it the ODSBox we have will not work

michaelsauter:
@braisvq1996 I think this just confirms that you'll need to investigate an alternative to the current ODS Box. Maintaining CentOS/RHEL7 base images will be a hassle more and more, plus the box is OpenShift 3.11 only so won't give you reliable results. The box should not block the uptake of Node 18 I believe.

braisvq1996:
@michaelsauter you have your reason, the box is using OKD3, and this does not "help" for Openshift 4 tests....
Either way I would not break the box "on purpose" if we can avoid it

tbugfinder:
As per my understanding nodejs project will not build CentOS7 binaries.
https://github.com/nodejs/node/issues/43246

So for CentOS7 we could compile proper GLIBC version 2.27 within the Image and see if that would work with the RHEL8 binary.

In regards to the ubi8 image, do you want to wait for RHEL providing nodejs18 repo (sounds like Sept/Oct timeframe) or use the installer of the CentOS7 Dockerfile which would already be available.

tbugfinder:
@braisvq1996 @cschweikert 
NodeJS 18 + ubi8 built successfully.

IMHO, CentOS7 support has to dropped here to move forward.

braisvq1996:
@tbugfinder @cschweikert centos agents have been removed so if the conficts get resolved we could merge this into master

cschweikert:
@braisvq1996 thanks finishing this one ❤️

PR #803, opened 2022-05-10
add comment in test fixture as module name and value of parameter name have to match

### LINKED ISSUE #802: inf-terraform-aws & inf-terraform-azure: test-fixture document condition
It is not obvious that local.name and the module name have to be the same.
A single simple comment might ease developer experience.

https://github.com/opendevstack/ods-quickstarters/blob/5d29cd96976c4b96aa735f898c5f5c5796154b4e/inf-terraform-aws/files/test/fixtures/default/main.tf#L2

FYI: @nichtraunzer

tbugfinder:
add comment in test fixture as module name and value of parameter name have to match.
This should improve the development experience as the hidden condition is shared.

Closes #802 
Fixes #802 

Tasks: 

### DIFF
--- a/inf-terraform-aws/files/test/fixtures/default/main.tf
+++ b/inf-terraform-aws/files/test/fixtures/default/main.tf
@@ -8,6 +8,7 @@ locals {
 data "aws_region" "current" {}
 
 module "stack-aws-quickstarter-test" {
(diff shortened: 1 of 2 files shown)


nichtraunzer (approved):
lgtm

Issue #804, opened 2022-05-12
Jcenter repository not available and build step fails


braisvq1996:
**Describe the bug**
Currently Jcenter seems not available and have made the pipelines fails due to a package that was not stored properly in Nexus.
In the Java/SpringBoot QS we have this repository ([here](https://github.com/opendevstack/ods-quickstarters/blob/master/be-java-springboot/templates/gradle-repositories.template#L11)) and maven public below.
Maven public is a group type repository in Nexus and already contains[ jcenter and maven-central](https://github.com/opendevstack/ods-core/blob/master/nexus/json/createRepos.json#L24-L26) (that points to https://repo1.maven.org/maven2/) so it does not makes sense to have both in gradle

@renedupont

renedupont:
In that case I think we can remove [this](https://github.com/opendevstack/ods-core/blob/master/nexus/json/createRepos.json#L10) from the nexus repos?

braisvq1996:
yeah, we could remove it from Nexus and for the Quickstarters
At the end, we have maven-central for this kind of packages

victorpablosceruelo:
This is needed right now. Quickstarters cannot compile because jcenter dependencies cannot be found.

https://github.com/opendevstack/ods-quickstarters/pull/807/commits/6ae942407bdc53c752325fc057271b6ad6f2407a

Thanks for pointing it out, Brais.

victorpablosceruelo:
More info about this:
https://stackoverflow.com/questions/66651640/jcenter-deprecation-impact-on-gradle-and-android#:~:text=Summary%3A%20After%20February%201%2C%202022%20the%20jcenter%20%28%29,will%20be%20removed%20in%20the%20next%20major%20release.

victorpablosceruelo:
Provisioning app is affected by this too:
https://github.com/opendevstack/ods-provisioning-app/issues/737

PR #807, opened 2022-05-18
4.x <- Full revision of Jenkins Pipelines, to make them work again

victorpablosceruelo:
The upgrade of the atlassian stack needed some modifications in quickstarters because some bugs appeared.

Tasks: 
- [x] Remove jcenter repositories from quickstarters (Fixes https://github.com/opendevstack/ods-quickstarters/issues/804 ).
- [x] Fix non-working jdk-17 usage (Fixes https://github.com/opendevstack/ods-quickstarters/issues/808 ).
- [x] Full revision of Jenkins Pipelines, to make them work again. Increased timeouts for building quickstarters and added the retrieval of the return status for building each quickstarter.
- [x] UBI8 docker image for terraform is installing packages from centos when it should take the ones available at Red Hat (Fixes https://github.com/opendevstack/ods-quickstarters/issues/813 )
- [x] Upgrade atlassian stack (Implements https://github.com/opendevstack/ods-core/issues/1138 )

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - Switch fe-angular, fe-ionic and be-typescript-express to Node.js 16 builder agent ([#763](https://github.com/opendevstack/ods-quickstarters/issues/763)
 - Update and improve e2e-cypress quickstarter ([#770](https://github.com/opendevstack/ods-quickstarters/issues/770))
 - Update fe-ionic to Ionic 6.19.0 ([#780](https://github.com/opendevstack/ods-quickstarters/issues/780))
+- Upgrade atlassian stack (Implements [#1138](https://github.com/opendevstack/ods-core/issues/1138))
 
 ### Fixed
 
@@ -35,7 +36,9 @@
 - ODS AMI build failing due an E2E test error of ionic quickstarter ([#742](https://github.com/opendevstack/ods-quickstarters/issues/742))
 - ODS AMI build failing due an missing list of supported browsers in ionic quickstarter ([#756](https://github.com/opendevstack/ods-quickstarters/issues/756))
 - inf-terraform-aws: Fix error handling of Makefile ([#680](https://github.com/opendevstack/ods-quickstarters/issues/680))
-- Removes jcenter repositories and fixes jdk-17 problems. ([#807](https://github.com/opendevstack/ods-quickstarters/pull/807))
+- Remove jcenter repositories from quickstarters (Fixes [#804](https://github.com/opendevstack/ods-quickstarters/issues/804))
+- Fix non-working jdk-17 usage (Fixes [#808](https://github.com/opendevstack/ods-quickstarters/issues/808))
+- Full revision of Jenkins Pipelines, to make them work again. Increased timeouts for building quickstarters and added the retrieval of the return status for building each quickstarter.
 
 ## [4.0] - 2021-11-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@
 - ODS AMI build failing due an E2E test error of ionic quickstarter ([#742](https://github.com/opendevstack/ods-quickstarters/issues/742))
 - ODS AMI build failing due an missing list of supported browsers in ionic quickstarter ([#756](https://github.com/opendevstack/ods-quickstarters/issues/756))
 - inf-terraform-aws: Fix error handling of Makefile ([#680](https://github.com/opendevstack/ods-quickstarters/issues/680))
+- Removes jcenter repositories and jdk-17 usage ([#807](https://github.com/opendevstack/ods-quickstarters/pull/807))
 
 ## [4.0] - 2021-11-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@
 - ODS AMI build failing due an E2E test error of ionic quickstarter ([#742](https://github.com/opendevstack/ods-quickstarters/issues/742))
 - ODS AMI build failing due an missing list of supported browsers in ionic quickstarter ([#756](https://github.com/opendevstack/ods-quickstarters/issues/756))
 - inf-terraform-aws: Fix error handling of Makefile ([#680](https://github.com/opendevstack/ods-quickstarters/issues/680))
-- Removes jcenter repositories and jdk-17 usage ([#807](https://github.com/opendevstack/ods-quickstarters/pull/807))
+- Removes jcenter repositories and fixes jdk-17 problems. ([#807](https://github.com/opendevstack/ods-quickstarters/pull/807))
 
 ## [4.0] - 2021-11-05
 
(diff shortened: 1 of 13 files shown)


cschweikert (changes requested):
Hi everybody, I don't know the background of this change, but since it claims to remove usage of Java 17, the removal should rather be clean and complete. In `common/jenkins-agents/maven/docker/Dockerfile.centos7` I can still find `yum install ... java-17-...` and calling of `use-j17.sh`, which again I'd rather delete then breaking each line with `exit 1`. If java 17 should come back at a later point in time, there is still the git history, release tags, etc. with this code.

victorpablosceruelo:
> Hi everybody, I don't know the background of this change, but since it claims to remove usage of Java 17, the removal should rather be clean and complete. In `common/jenkins-agents/maven/docker/Dockerfile.centos7` I can still find `yum install ... java-17-...` and calling of `use-j17.sh`, which again I'd rather delete then breaking each line with `exit 1`. If java 17 should come back at a later point in time, there is still the git history, release tags, etc. with this code.

Hi.
I think you're right. Since we have the following bug in ods-core, we should remove usage of jdk 17 completely.

https://github.com/opendevstack/ods-quickstarters/issues/808

I'll do it asap.
Thanks !!

victorpablosceruelo:
I'm referencing this issue ( jcenter repositories no more working) reported by Brais, so we know where we are taking care of fixing it.

https://github.com/opendevstack/ods-quickstarters/issues/804

victorpablosceruelo:
> Hi everybody, I don't know the background of this change, but since it claims to remove usage of Java 17, the removal should rather be clean and complete. In `common/jenkins-agents/maven/docker/Dockerfile.centos7` I can still find `yum install ... java-17-...` and calling of `use-j17.sh`, which again I'd rather delete then breaking each line with `exit 1`. If java 17 should come back at a later point in time, there is still the git history, release tags, etc. with this code.

Seeing in detail usages of use-j17.sh, we have 2 docker images: 

https://github.com/opendevstack/ods-quickstarters/blob/task/upgrade-atlassian-stack/common/jenkins-agents/maven/docker/Dockerfile.centos7
https://github.com/opendevstack/ods-quickstarters/blob/task/upgrade-atlassian-stack/common/jenkins-agents/maven/docker/Dockerfile.ubi8

while in the centos image the usage of jdk17 does not work because of https://github.com/opendevstack/ods-quickstarters/issues/808 , it might work in the ubi image. 

So, I think the best solution right now would be letting the script work again (so it can be used in ubi image) 
https://github.com/opendevstack/ods-quickstarters/pull/807/files#diff-7a65b51f77e65d2dddd264e79341f008fde7d052ba147a27115cc78b72e16d4e
https://github.com/opendevstack/ods-quickstarters/pull/807/files#diff-b81ace78800b2647fd42dae0eb5525973780a175c369ad57fb4656aad5139f3d
but 
avoiding testing that it works during centos 7 img generation 
https://github.com/opendevstack/ods-quickstarters/pull/807/files#diff-476fb2effd01084c0826e60a751d73637841c0c7a7dc870591080a71c6efc548

and not using it yet to build quickstarter
https://github.com/opendevstack/ods-quickstarters/pull/807/files#diff-3a1a11113a04ba34d1dfdbda57d6f46b2ba4ceae7ad133c078b1755abf646fbe

I think it is the best solution. What do you think?

cschweikert:
> while in the centos image the usage of jdk17 does not work because of #808 , it might work in the ubi image.

Still I'm wondering, if we can then get rid of `java-17-openjdk-...` (yum install) and `use-j17.sh` (chmod) in the centos thingy. Just falls into my eye.

@stitakis How about the rest of this PR?

cschweikert:
> while in the centos image the usage of jdk17 does not work because of #808 , it might work in the ubi image.

Still I'm wondering, if we can then get rid of `java-17-openjdk-...` (yum install) and `use-j17.sh` (chmod) in the centos thingy. Just falls into my eye.

@stitakis How about the rest of this PR?

cschweikert:
@victorpablosceruelo It seems that you are still developing on this branch. I guess you try to get the CI tests running. Have you tried testing those builds related to your changes locally? In general this should be faster than commiting to github each time. Also think about the firework of builds (and mails 😉) that are getting triggered with each commit.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@
 - ODS AMI build failing due an E2E test error of ionic quickstarter ([#742](https://github.com/opendevstack/ods-quickstarters/issues/742))
 - ODS AMI build failing due an missing list of supported browsers in ionic quickstarter ([#756](https://github.com/opendevstack/ods-quickstarters/issues/756))
 - inf-terraform-aws: Fix error handling of Makefile ([#680](https://github.com/opendevstack/ods-quickstarters/issues/680))
+- Removes jcenter repositories and jdk-17 usage ([#807](https://github.com/opendevstack/ods-quickstarters/pull/807))
```
cschweikert, line 38: isn't this also a breaking change? If so I'd move this entry to `### Modified`. Also I'd rename to `- Remove JCenter repositories and jdk 17 support ...` or so

stitakis:
Hi all, I not agree on removing JDK 17 support without first exploring a way to solve the idk 17 installation problem in centos 7. Was this tried out? Another observation, the title of this ticket does not match to the changes. This shouldn't be the case.

stitakis (changes requested):
Before reverting the support of jdk 17 I suggest to try to solve the installation problem in centos first.

cschweikert (dismissed):
I reviewed some of the changes again. Unfortunatelly I'm having problems understanding the reasoning of some of these changes (e.g. removal of jcenter). Also some of the changes don't seem to be related to the initial description of the PR (e.g. about nodejs12 and terraform). For those it might make sense to move them over into a new issue+PR. Also it would be good to have an issue that this PR is directly based on (e.g. "Full revision of Jenkins Pipelines, to make them work again" seems to point to an issue).

Thanks @victorpablosceruelo for all the work you put into this PR. Seems like you needed to try a lot of crazy stuff to get it running.

victorpablosceruelo:
Hi @cschweikert 
Thanks for your comments.
You are right. The PR title is not valid anymore. I changed it.
Besides, I added extra information on what are the bugs or issues we achieve with this PR.
Hope you like it a little bit more now
Thanks !!!

CHANGELOG.md
victorpablosceruelo: I've changed it. It was an initial message. Now it is better explained.
Do you see it ok now?

victorpablosceruelo: Yes. You're right. It was a breaking change.
But we found a different solution that keeps jdk 17, using Adoptium Temurin, maintained by Eclipse.
More information in the following link:
https://github.com/opendevstack/ods-quickstarters/issues/808

cschweikert (approved):
@victorpablosceruelo THX again for bringing this complex topic forward 😀

victorpablosceruelo:
Thanks @cschweikert  for your comments. 

I'm getting the following error in GitHub:
The base branch restricts merging to authorized users.
(https://docs.github.com/articles/about-protected-branches/)

I do not Know what to do at this point...

cschweikert:
yeah, `4.x` is a protected branch that only project admins can merge into. Maybe @michaelsauter or @braisvq1996 can help here 😊

PR #925, opened 2023-05-31
Always pull image by default

### LINKED ISSUE #874: imagePullPolicy: IfNotPresent may lead to image verification errors
**Describe the bug**
[Current imagePullPolicy for all components: "ifNotPresent"](https://github.com/opendevstack/ods-quickstarters/blob/4.x/common/ocp-config/component-environment/component-template.yml#L91)

We believe that the current configuration can lead to inconsistent results (observed already on passes to prod): when it comes to verifying the image of the running container, we have seen it already failing in a couple of instances:

[Image verification](https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/src/org/ods/services/OpenShiftService.groovy#L586)

**To Reproduce**
It does not happen everytime, but the steps would roughly be:
1. Prepare to deploy to prod
2. Observe that the imagestreams get imported
3. Rollout reported as succesfull
4. Image verification fails (reporting that the running container is using an old (and sometimes no longer existing in the namespace's Imagestream list) ImageID

We believe this might be caused by _some nodes_ not correctly pulling the image, maybe considering it already present.

**Expected behavior**
Images should be always pulled to ensure integrity

**Screenshots**

 - OpenShift: 4.10
 - OpenDevStack 4.0
 
**Log Output**
When the error occurs, the log will output something like
```
Container '${containerName}' is using image '${runningImageSha}' which differs from defined image '${definedImageSha}'.

```

To summarize: we think it would be beneficial to switch from "imagePullPolicy: IfNotPresent" to "imagePullPolicy: always" to avoid these inconsistencies, even when this could potentially impact performance.

braisvq1996:
Closes #874

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Update of Maven agent, Java and Spock/Geb quickstarter ([#878](https://github.com/opendevstack/ods-quickstarters/issues/878))
 - inf-terraform-[aws|azure], add new jenkins-agent-terraform-2306 with updated tools (ruby 3.2.2, python 3.11, etc.) and dependencies, add tflint, mark other jenkins-agent-terraform as deprecated([#914](https://github.com/opendevstack/ods-quickstarters/issues/914))
 - Update of Python agent, Python, Streamlit and Jupyter quickstarters ([#902](https://github.com/opendevstack/ods-quickstarters/issues/902))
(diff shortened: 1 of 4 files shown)


michaelsauter (approved):
As discussed, I dislike both changes:

* instead of pulling latest all the time, it would be better to use "immutable" image tags
* recreate introduces downtime which is not necessary

But I get why you are making the change, so I'll approve.

PR #934, opened 2023-06-22
inf-terraform-aws: remove cloudformation stack from default quickstarter

nichtraunzer:
We experienced that many users forget to remove the Cloudformation stack and its corresponding inspec test. This usually causes issues. Therefore we decided to replace that with a simple region test.

Tasks: 


tbugfinder (approved):
LGTM&LGTY

PR #939, opened 2023-07-04
Fix Jenkinsfile params in StreamLit

braisvq1996:
Fix JenkinsFile params for Streamlit to align with the rest of the quickstarters


gerardcl (approved):
LGTM

PR #946, opened 2023-08-23
add chmod g=rwx $HOME to jenkins agent (#945)

### LINKED ISSUE #945: nodejs jenkins-agent can not import private keys into gpg keyring to use with helm secrets
**Describe the bug**
I was using the nodejs16 jenkins-agent with helm to deploy my application. The deployment failed with the following error message (while trying to load the private key into the gpg keyring):
```bash
gpg: WARNING: unsafe ownership on homedir '/home/jenkins/.gnupg'
gpg: failed to create temporary file '/home/jenkins/.gnupg/.#l...': Permission denied
gpg: keyblock resource '/home/jenkins/.gnupg/pubring.kbx': Permission denied
gpg: no writable keyring found: Not found
gpg: key ****: failed to re-lookup public key: No public key
gpg: error reading ****: No public key
gpg: import from **** failed: No public key
gpg: Total number processed: 0
gpg:       secret keys read: 1
```

**Workaround**
Switching to the base jenkins-agent and use containerTemplates for stages in which you need nodejs.

**To Reproduce**
1. Start a pipeline that uses helm secrets and any nodejs jenkins-agent

**Expected behavior**
helm secrets should work with the nodejs jenkins-agents

 - OpenShift: 4.x
 - OpenDevStack 4.x

**Additional context**
Further debugging revealed, that in the base jenkins-agent there is no `/home/jenkins/.gnupg` folder existing in the image (on pipeline start). See Screenshot 1. This works fine as the user used in the pipeline is in the root group and has write permissions on this folder. That way the user can create the `.gnupg` folder and the keyring in it, the first time gpg is used.
In the nodejs16 jenkins-agent there is already a `/home/jenkins/.gnupg` folder existing in the image (on pipeline start). As this folder was created during the `docker build`, this folder is owned by the user used during the build and afterwards handed over to user 1001 ([in L59 in the Dockerfile](https://github.com/opendevstack/ods-quickstarters/blob/master/common/jenkins-agents/nodejs16/docker/Dockerfile.ubi8#L59)). See Screenshot 2. If this image is now used in the pipeline the pipeline will run with a different user (no the `USER 1001` directive does not change anything as 1001 is not in the user id range that OpenShift is using (See [RedHat KB](https://www.redhat.com/sysadmin/linux-file-permissions-explained#:~:text=Execute%20(x),-This%20permission%20is&text=Having%20execute%20permission%20on%20a,way%20to%20a%20subdirectory%20underneath.))). Therefore this user does not have the x permission on this folder and can not access files in it, including the keyring file.
I looked also into the other nodejs jenkins-agents: They seem to have the same issue (I don't know if it is the same for other jenkins-agents as well)
I looked into what is causing the creation of the `.gnupg` folder on image build. At least for the nodejs16 agent it is installing yarn ([L36](https://github.com/opendevstack/ods-quickstarters/blob/master/common/jenkins-agents/nodejs16/docker/Dockerfile.ubi8#L36)) and google-chrome ([L44](https://github.com/opendevstack/ods-quickstarters/blob/master/common/jenkins-agents/nodejs16/docker/Dockerfile.ubi8#L44)).
For the yarn part, this can be fixed by switching to the recommended installation methods from yarn ([legacy](https://classic.yarnpkg.com/en/docs/install#windows-stable) or the [recommended](https://yarnpkg.com/getting-started/install)). For the google-chrome part I don't have any idea yet. Maybe the easier fix would be to delete the `.gnupg` folder as last step in the image builds.

**Screenshots**
Screenshot 1
Screenshot 1
Screenshot 2
Screenshot 2023-08-22 091034

matzehecht:
this fixes the issue that the user used during pipeline run cannot access the child direcories created the image build. See #945 

Adds/Switches to `g=rwx` for affected jenkins-agents

Closes #945
Fixes #945

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Modified
 
 ### Fixed
+- jenkins agents can not import private keys into gpg keyring to use with helm secrets ([#945](https://github.com/opendevstack/ods-quickstarters/issues/945))
 
 ## [4.3.0] - 2023-07-13
 
--- a/common/jenkins-agents/jdk/docker/Dockerfile.ubi8
+++ b/common/jenkins-agents/jdk/docker/Dockerfile.ubi8
@@ -85,7 +85,7 @@ RUN mkdir -p $GRADLE_USER_HOME && \
 RUN bash -l -c 'echo export JAVA_OPTS="$(/tmp/set_java_proxy.sh && echo $JAVA_OPTS)" >> /etc/bash.bashrc'
 
 RUN chown -R 1001:0 $HOME && \
-    chmod -R g+rw $HOME && \
+    chmod -R g+rwX $HOME && \
     chmod -c 666 /etc/pki/ca-trust/extracted/java/cacerts && \
     ls -la /etc/pki/ca-trust/extracted/java/cacerts
 
--- a/common/jenkins-agents/nodejs12/docker/Dockerfile.ubi8
+++ b/common/jenkins-agents/nodejs12/docker/Dockerfile.ubi8
@@ -73,6 +73,6 @@ RUN npm config set registry=$nexusUrl/repository/npmjs/ && \
     echo yarn version: $(yarn --version)
 
 RUN chown -R 1001:0 $HOME && \
-    chmod -R g+rw $HOME
+    chmod -R g+rwX $HOME
 
 USER 1001
--- a/common/jenkins-agents/nodejs16/docker/Dockerfile.ubi8
+++ b/common/jenkins-agents/nodejs16/docker/Dockerfile.ubi8
@@ -57,6 +57,6 @@ RUN npm config set registry=$nexusUrl/repository/npmjs/ && \
     echo yarn version: $(yarn --version)
 
 RUN chown -R 1001:0 $HOME && \
-    chmod -R g+rw $HOME
+    chmod -R g+rwX $HOME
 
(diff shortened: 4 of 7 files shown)


paulmayer:
For potential consideration:

`g=rwX` could be an option to limit adding +x to directories and where a user already has x.

matzehecht:
> For potential consideration:
> 
> `g=rwX` could be an option to limit adding +x to directories and where a user already has x.

You are right. Changed it.

braisvq1996:
@matzehecht is this PR also related to https://github.com/opendevstack/ods-jenkins-shared-library/issues/1028 ?

matzehecht:
> @matzehecht is this PR also related to [opendevstack/ods-jenkins-shared-library#1028](https://github.com/opendevstack/ods-jenkins-shared-library/issues/1028) ?

No. Those are completely separate issues. As stated in the description: This PR solves the issue #945. In the description of the PR I posted a very detailed the description, which should make the difference clear.
But in short: This PR and #945 tackle an issue while importing a private key into the gpg keyring only on some jenkins agents (mainly the nodejs agents). The issue you mentioned tackles a problem that helm secrets sometimes can not find the right key in the keyring on a agent that is not affected by the problem described in #945.

braisvq1996 (approved):

cschweikert (approved):

Issue #962, opened 2023-10-24
Maintenance of Angular, Ionic, TypeScript and Cypress quickstarters plus new node version


braisvq1996:
There is a new release node LTS (node 20) so we should provide this new agent and update the quickstarters to use this new agent by default.

- [x] https://github.com/opendevstack/ods-quickstarters/pull/970
- [x]  Remove nodejs12 #977
- [x] Update and migrate angular QS to nodejs 20 https://github.com/opendevstack/ods-quickstarters/pull/972
- [x] Update and migrate ionic QS to nodejs 20 https://github.com/opendevstack/ods-quickstarters/pull/972
- [x] Update and migrate typescript QS to nodejs 20 https://github.com/opendevstack/ods-quickstarters/pull/972
- [x] Update and migrate cypress QS to nodejs 20 https://github.com/opendevstack/ods-quickstarters/pull/972

**Nodejs 20 agent (new)**

The comand from [nodesource](https://github.com/nodesource/distributions#nodejs-v20x) seems to be working in order to install node in the agent

**Angular Quickstarter**

| Package | Current Version | Sugested version | Location |
| ------------- | ------------- | ------------- | ------------- |
| nginx base image | 1.24.0-alpine  | 1.25.3-alpine | [Dockerfile](https://github.com/opendevstack/ods-quickstarters/blob/master/fe-angular/files/docker/Dockerfile#L1) |
| AngularCli | 15.2.4  | 17.0.3 | [Jenkinsfile](https://github.com/opendevstack/ods-quickstarters/blob/master/fe-angular/Jenkinsfile#L17) |

**Ionic Quickstarter**

| Package | Current Version | Sugested version | Location |
| ------------- | ------------- | ------------- | ------------- |
| nginx base image | 1.24.0-alpine  | 1.25.3-alpine | [Dockerfile](https://github.com/opendevstack/ods-quickstarters/blob/master/fe-ionic/files/docker/Dockerfile#L1) |
| IonicCli | 7.1.1  | 7.1.5 | [Jenkinsfile](https://github.com/opendevstack/ods-quickstarters/blob/master/fe-ionic/Jenkinsfile#L17) |

**TypeScript Quickstarter**

| Package | Current Version | Sugested version | Location |
| ------------- | ------------- | ------------- | ------------- |
[...]

PR #977, opened 2023-12-11
Remove nodejs12

### LINKED ISSUE #936: Remove nodejs12 agent
Remove nodejs 12 agent from the suported stack

braisvq1996:
Closes #936
Related to #962


cschweikert (changes requested):
I found another reference to NodeJS 12 in the docs: https://github.com/opendevstack/ods-quickstarters/blob/master/docs/modules/quickstarters/nav.adoc?plain=1#L24

And maybe there is more. I found this for example: https://github.com/opendevstack/ods-core/blob/0e02f3466cdd6f6e80a2c3a7d45509ff72f0ca97/tests/ods-verify/verify-ods-namespace-images_test.go#L24
That one is funny, because it misses any other node version.

braisvq1996:
Updated the docs in the quickstarter repository.
The tests on core need some work true. Next year we want to use the golden tests on every quickstarter in order to do the testing internally so this will not be abandoned.

cschweikert (approved):

Issue #1013, opened 2024-04-30
Helm ready Quickstarters - sunsetting tailor


gerardcl:
**Is your feature request related to a problem? Please describe.**
[Tailor](https://github.com/opendevstack/tailor) is a self-developed component of OpenDevStack (ODS) whose active development has ceased and is logically superseded by Helm.
As a user I require having control of the k8s resources I need for my components/deployments (actively or passively). 
As a support team member of ODS I can give better/faster support if resources are defined as code, in the same repository.

**Describe the solution you'd like**
We want to provide a base chart in the most generic ODS quickstarter component, in the `docker-plain`. This will be the baseline for the other ODS quickstarters and for existing components in existing running ODS installations.

We want to provide such baseline with proper documentation of how to use common resources: Environment vars, config maps, secrets, storage, replicas, resource requirements, metadata, network policies, cert-manager, ...

We want to transition from the deprecated DeploymentConfig resource type to Deployment.

We want to provide a change that is not a hard-cut from tailor but provides a transition process and proper sunsetting of if.

**Describe alternatives you've considered**
N/A

**Additional context**
When looking at the different user types of ODS, we will benefit from Helm charts, as it:
- pushes usage of IaC with already backed-in good practices and good defaults
- helps on faster support from platforms team
- empowers users to  self-configure/fix by proper FAQs/HowTos and because of using a industry standard (Helm), that has great documentation.

PR #1035, opened 2024-08-02
Docker plain helm chart

segfault16:
Add helm chart for docker plain

Refers to #1013 

Tasks: 
- [ ] Shared library adaptions


docker-plain/files/chart/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: "{{ include "chart.fullname" . }}-test-connection"
+  labels:
+    {{- include "chart.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": test
+spec:
+  containers:
+    - name: wget
+      image: public.ecr.aws/docker/library/busybox
```
renedupont, line 12: Would it make sense to let this point to our BI docker repo proxy instead of aws? When is this test executed? On every helm deployment?

docker-plain/files/chart/templates/_affinity.tpl
```diff
@@ -0,0 +1,51 @@
...
+
+Usage: Include where needed, e.g.
+````
+apiVersion: apps/v1
+kind: Deployment
+spec:
+  template:
+    spec:
+      affinity:
+        podAntiAffinity: {{- include "common.affinities.pods.soft" . | nindent 10}}
```
renedupont, line 19: Is there a specific reason for having _affinity.tpl and _labels.tpl as separate files even though there definition both start with 'common'? Would it make sense to put them together into a _common.tpl?

docker-plain/files/chart/templates/deployment.yaml
```diff
@@ -0,0 +1,70 @@
...
+  labels:
+    {{- include "chart.labels" . | nindent 4 }}
+spec:
+  {{- if not .Values.autoscaling.enabled }}
+  replicas: {{ .Values.replicaCount }}
+  {{- end }}
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 0%
```
renedupont, line 14: is the values of maxUnavailable and maxSurge some best practices we do at BI or how come these are "hardcoded" here? Or are these defaults?

docker-plain/files/chart/templates/tests/test-connection.yaml
renedupont: oh wait, we can't put our inhouse URL here in github :)

docker-plain/files/chart/templates/_affinity.tpl
gerardcl: IMHO I would keep it as proposed, as this is based on good practise templating in helm (based on the concept in coding in general, to split code/project source by context), in one template we handle affinity in another labels and so on. This helps keep complexity low.

docker-plain/files/chart/templates/deployment.yaml
gerardcl: As per [docs](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#rolling-update-deployment), the defaults for both rolling update config params is 25%.
I would make it configurable too, I will provide a new commit for this! good finding @renedupont! 💯

gerardcl: @renedupont  @segfault16  please take a look at the new commit https://github.com/opendevstack/ods-quickstarters/pull/1035/commits/0c59afcd5a180ce30200d6856c65a65be8320591 that faces this comment and give your feedback, thanks!

docker-plain/files/chart/templates/_affinity.tpl
renedupont: Hehe that was actually my point as well when I proposed the change. IMO, currently, there is a _helpers.tpl (not good splitted/named by context btw. 😆) in which the definitions all start with 'chart'. So I see 'chart' as the context here. For the definitions inside _affinity.tpl as well as _labels.tpl they all start with 'common', so I guess the context here is common?

docker-plain/files/chart/templates/deployment.yaml
renedupont: Do we specifically need to put both values in the values.yaml if we set them to the default values anyway? Are they that important or often changed to put them there? I am just thinking the more stuff we put in there, the more complex it looks like (especially for kubernetes/helm beginners).

gerardcl: Makes sense, I made the proposal to move that there as we want users be enabled on configuration capabilities via values.yaml. So, to having as a goal to (almost) only use values.yaml and require minimal touch (if none at all) on the resources yamls.
I would either then do this way already proposed, or make it more elaborated logic in the deployment yaml like: if you want recreate then it just writes down the recreate statement and if you want rollingupdate then it writes down the rollingupdate statements and the configured/defined required values.
what would you do?
Unless you both prefer to just keep what it was done already and see how this is used later on and, if required, improve from feedback once released. And maybe only do a tiny improvement by pointing to docs and why we propose such values.

renedupont: As just discussed: Lets keep the fields, but don't use the defaults but the proposed values (50% and 0%)

gerardcl:
> The base image has been changed to `nginxinc/nginx-unprivileged` because the alpine one did not work properly with the container probes

and this also now does match with the values.yaml image section, perfect!

BraisVQ:
Provision, build, and release pipelines have been tested and work properly with the latest commits out of the box

docker-plain/files/chart/DEVNOTES.md
```diff
@@ -0,0 +1,41 @@
+# TODO
```
BraisVQ, line 1: Just in case we forget, this file should be removed before the PR is merged

docker-plain/files/docker/Dockerfile
```diff
@@ -1,12 +1,10 @@
 # add /overwrite FROM with your base image, and do whatever you like here :)
-FROM alpine:latest
+FROM nginxinc/nginx-unprivileged:latest
```
segfault16, line 2: - [ ]  Potential docker.io rate limit issue

gerardcl: good point! I will provide a solution based on using an optional proxy repository

docker-plain/files/chart/DEVNOTES.md
```diff
@@ -0,0 +1,41 @@
...
+- creation of template files folder in common, so we try to centralise the chart creation and maintenance from one place (as it is done already with openshift templates/tailor)
+
+Decisions:
+- To stay close to default helm templates: Remove the Values.componentId and use chart.fullname instead -> Otherwise breaks DEV + PREVIEW. Chart.Name should be the source for componentId. If we want to automate -> template the Chart.yaml on provisioning <- DONE
+
+Shared library ToDo's:
+- Match required labels (app)
+- Review helm install command
+- Release manager image checks
+- Agree if helm install requires overwritting values on image (see helm notes and deployment, and values.yaml)
```
segfault16, line 40: We update the readme to show how you can replicate jenkins behaviour locally, e.g.:

docker build -t $(git rev-parse --short) ...
helm upgrade --install --atomic --set imageId=$(git rev-parse --short) ...

docker-plain/files/chart/DEVNOTES.md
```diff
@@ -0,0 +1,41 @@
...
+- provide docs on how to test chart updates, by running `helm --debug template . ` under the chart folder to be tested
```
segfault16, line 41: also helm lint

docker-plain/files/chart/DEVNOTES.md
```diff
@@ -0,0 +1,41 @@
...
+        - future add as a Jenkins shared lib stage
+        - now either we do it as an added manually in the jenkinsfiles or documented and test locally
+        - make use of it on the github actions when PRs on quickstarters (reuse max as possible)
+- we will move to values.yaml
+    - [X] the probes
+    - [x] the affinity (missing labels now and some additions required)
+    - [x] the route (with timeout values and example for ACME usage)
+    - [x] rolling update strategy?
+- [X] remove provisioning resources creation -> get rid odsQuickstarterStageCreateOpenShiftResources
+- jenkinsfile with values.env.yaml ready, so we will provide all the env values too: Different replicaCount + ingress hosts
```
segfault16, line 15: add all 3 values.env.yaml

docker-plain/files/docker/Dockerfile
```diff
@@ -1,10 +1,6 @@
-# add /overwrite FROM with your base image, and do whatever you like here :)
-FROM nginxinc/nginx-unprivileged:latest
+FROM registry.access.redhat.com/ubi9/nginx-122
```
BraisVQ, line 1: we could use ubi9/nginx-124 instead

docker-plain/files/README.md
```diff
@@ -1,5 +1,46 @@
...
+
+Using a local kubernetes cluster you can deploy the component:
+
+```bash
+kubectl create ns docker-plain
+kind load docker-image testing/my-component:$(git rev-parse --short=8 HEAD)
+helm upgrade --install --wait --atomic --namespace docker-plain --set image.path=testing --set image.name=my-component --set image.tag=$(git rev-parse --short=8 HEAD) docker-plain chart
+```
+
 ## How to create a custom jenkins-agent out of this docker-plain component
```
segfault16, line 44: Still needed?

docker-plain/files/chart/values.yaml
```diff
@@ -0,0 +1,128 @@
...
+  #   paths:
+  #     - path: /
+  #       pathType: Prefix
+  tls: [] # If `tls` is left empty then the default OpenShift TLS config will be loaded (i.e.: TLS edge termination with HTTP redirect to HTTPS)
+  # - secretName: chart-example-tls
+  #   hosts:
+  #     - yourapp.yourdomain.com
+
+service:
+  enabled: true
```
renedupont, line 47: service.enabled is never used

faust2199:
Having a very strict values.schema.json is probably not going to provide good experience when using helmValues defined from the Jenkinsfile. According to https://helm.sh/docs/topics/charts/ , "the schema is applied to the final .Values object, and not just to the values.yaml file."

Let's say we have the following in the Jenkinsfile:
`
  odsComponentStageRolloutOpenShiftDeployment(context, [
    'selector': "app.kubernetes.io/name=${context.componentId}",
    'helmEnvBasedValuesFiles': ['values.env.yaml'],
    'helmValues': [
      'foo': 'bar'
    ]
  ]
`

When the above runs on Jenkins, ods-jenkins-shared-library will add two arguments on the command line:
`--set foo=bar --set global.foo=bar`

Then it will be aborted because additionalProperties are not allowed in both places. Is the user really expected to add two entries in values.schema.json to fix this?

I suggest that we remove the whole values.schema.json file. If the result of rendering is bad, `helm upgrade` is going to fail and rollback anyway.

renedupont:
> Having a very strict values.schema.json is probably not going to provide good experience when using helmValues defined from the Jenkinsfile. According to https://helm.sh/docs/topics/charts/ , "the schema is applied to the final .Values object, and not just to the values.yaml file."
> 
> Let's say we have the following in the Jenkinsfile: `odsComponentStageRolloutOpenShiftDeployment(context, [ 'selector': "app.kubernetes.io/name=${context.componentId}", 'helmEnvBasedValuesFiles': ['values.env.yaml'], 'helmValues': [ 'foo': 'bar' ] ]`
> 
> When the above runs on Jenkins, ods-jenkins-shared-library will add two arguments on the command line: `--set foo=bar --set global.foo=bar`
> 
> Then it will be aborted because additionalProperties are not allowed in both places. Is the user really expected to add two entries in values.schema.json to fix this?
> 
> I suggest that we remove the whole values.schema.json file. If the result of rendering is bad, `helm upgrade` is going to fail and rollback anyway.

I partially agree, I think there is a middle way though. In the `values.schema.json` we could set [`"additionalProperties": false` ](https://json-schema.org/understanding-json-schema/reference/object#additionalproperties) to `true` in the relevant parts. That way we validate everything that is set in that file and everything else will not be considered and not lead to failing the helm validation.

docker-plain/files/chart/values.yaml
```diff
@@ -0,0 +1,128 @@
...
+  requests:
+    cpu: 50m
+    memory: 64Mi
+
+autoscaling:
+  enabled: false
+  minReplicas: 1
+  maxReplicas: 100
+  targetCPUUtilizationPercentage: 80
+  # targetMemoryUtilizationPercentage: 80
```
faust2199, line 117: Enabling autoscaling in `values.yaml` doesn't work yet, without the corresponding HPA resource

If a user sets `enabled` to true, currently the behavior is `replicas` will be unset in `deployment.yaml`, it will always have exactly 1 replica, which is the default.

I suggest we either remove this section in values.yaml, or add HorizontalPodAutoscaler resource to make autoscaling work as expected.

By the way, having 100 as the default value of `maxReplicas` seems a bit too many to me. Maybe something below 10 would be good enough?

docker-plain/files/chart/values.yaml
faust2199: Fixed with commit fe5fa7f

docker-plain/files/README.md
jafarre-bi: Any answer?

docker-plain/files/chart/values.yaml
```diff
@@ -0,0 +1,127 @@
...
+    cpu: 100m
+    memory: 32Mi
+  requests:
+    cpu: 10m
+    memory: 16Mi
+
+autoscaling:
+  enabled: false
+  minReplicas: 1
+  maxReplicas: 3
```
jafarre-bi, line 115: The Release Manager doesn't support multiple replicas yet. Not sure if these values should even be configurable, or at least maxReplicas should be 1 by default, with a comment saying that higher values aren't supported.

jafarre-bi (changes requested):
A few new comments and some conversations still open.

docker-plain/files/chart/values.yaml
segfault16: @jafarre-bi I don't know about this limitation, there's nothing in opendevstack documentation about. Please provide the source

docker-plain/files/chart/DEVNOTES.md
segfault16: removed

docker-plain/files/README.md
segfault16: we'll keep it

docker-plain/Jenkinsfile.template
```diff
@@ -20,7 +20,10 @@ odsComponentPipeline(
     */
     odsComponentStageBuildOpenShiftImage(context)
   }
-  odsComponentStageRolloutOpenShiftDeployment(context)
+  odsComponentStageRolloutOpenShiftDeployment(context, [
+    'selector': "app.kubernetes.io/name=${context.componentId}",
```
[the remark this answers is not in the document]
faust2199: > Why is this selector only the name? chart.selectorLables are, as they should, name + instance.

Do you mean https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-instance? How do we generate the unique postfix like in the example given "mysql-**abcxyz**"?

docker-plain/files/chart/values.yaml
jafarre-bi: HelmDeploymentStrategy, line 198:
// TODO: Once the orchestration pipeline can deal with multiple replicas,
// update this to store multiple pod artifacts.
Even if the code could deal with it and the documentation and traceability would be correctly preserved, it's still not officially supported. We would need to validate the functionality, document it and create specific automated tests to prevent regressions.

docker-plain/Jenkinsfile.template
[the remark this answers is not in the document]
jafarre-bi: My questions here are:
1. Is the chart name always equal to the componentId?
2. Will the functionality of getting resources based only on the name label never return resources that shouldn't match?
3. Is it correct to specify this selector in the deployment mean when the actual selector used is different? (For this I would answer no)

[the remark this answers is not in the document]
jafarre-bi: As agreed in the meeting, we will make the helmReleaseName config parameter explicit and add it to the selector to match the actual selector used in the deployments.

docker-plain/files/chart/values.yaml
jafarre-bi: As agreed, I will see if we are on time to add a note stating that multiple replicas are currently unsupported to the official documentation. We will add a comment to this values file referencing that to warn the users to use this at their own risk.

jafarre-bi (commented):
Adding comments to the open conversations with the outcome of the meeting we just had.

jafarre-bi:
> As agreed in the meeting, we will make the helmReleaseName config parameter explicit and add it to the selector to match the actual selector used in the deployments.

Has this open point been addressed?
If we can close the open points in the next few days, we aim to release the Helm quickstarter in the next EDP release, which will start mid-November.
Please, @faust2199 , @segfault16, and everyone else, we are almost there —let's work together to close this.
Thank you!

jafarre-bi (changes requested):
@segfault16, please see the comments to the latest changes.

docker-plain/Jenkinsfile.template
```diff
@@ -20,9 +20,12 @@ odsComponentPipeline(
     */
     odsComponentStageBuildOpenShiftImage(context)
   }
+  def releaseName = context.componentId // can be customized as needed
```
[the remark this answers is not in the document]
jafarre-bi: @segfault16 The deployment.yaml template uses chart.selectorLabels for its selector. In _helpers.yaml, chart.selectorLabels contains an instance label with the hardcoded value `{{ .Release.Name }}`, which matches Helm recommendations. Saying "can be customised as needed" may induce people to think any value is fine there. Saying "must match .Release.Name" would make things much clearer for the readers, independently of the RM providing the right value when updating.

jafarre-bi (approved):
Congratulations everybody!!

PR #1044, opened 2024-08-20
Adds readme doc about gradle cache in pipeline

### LINKED ISSUE #1043: Add doc about adding gradle-cache to build pipeline
One can improve the build pipeline time by adding a Gradle cache. This way, `gradlew` can use 
previously downloaded distributions instead of downloading them from the internet on every build.

A how to guide about this could be added to the readme. 

I provided a PR for this: #1044

renedupont:
Added readme file to springboot QS and added section for using gradle cache in the pipeline

Closes #1043

### DIFF
--- a/be-java-springboot/files/README.md
+++ b/be-java-springboot/files/README.md
@@ -1,0 +1,59 @@
+# {{project-name}} - provisioned from the OpenDevStack Java/Spring Boot QuickStarter (be-java-springboot)
+
+The official OpenDevStack documentation for this QuickStarter can be found [here](https://www.opendevstack.org/ods-documentation/opendevstack/latest/quickstarters/be-java-springboot.html).
+
+### Adding Gradle cache to your CI/CD pipeline
+
+One can improve the build pipeline time by adding a Gradle cache. This way, `gradlew` can use 
+previously downloaded distributions instead of downloading them from the internet on every build.
+To do so, follow the steps below:
+
+- Create a `PersistentVolumeClaim` (PVC) with the name `gradle-cache` in your `{{project-name}}-cd`-namespace,
+either via the OpenShift UI or via `oc` like this:
+```
+oc create -n {{project-name}}-cd -f - <<EOF
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+name: gradle-cache
+spec:
+accessModes:
+- ReadWriteOnce
+resources:
+requests:
+storage: 2Gi
+EOF
+```
+The above needs to be done only once per project, as this PVC can be reused for other components within the same project.
+
+- In your `Jenkinsfile`, mount this PVC to the containers of the pipeline by adding it as a podVolume:
+```
+...
+odsComponentPipeline(
+  podContainers: [
+    ...
+  ],
+  podVolumes: [
+    persistentVolumeClaim(claimName: "gradle-cache", mountPath: "/.gradle-cache", readOnly: false)
+  ],
+  branchToEnvironmentMapping: [
+    ...
+  ]
+)
+...
+```
+
+- Add this mount path as an env var called `GRADLE_USER_HOME` in the build stage to be picked up by Gradle:
+```
+...
+def stageBuild(def context) {
+  stage('Build and Unit Test') {
+    withEnv(["TAGVERSION=${context.tagversion}",...,"GRADLE_USER_HOME=/.gradle-cache",...]) {
+      ...
+    }
+  }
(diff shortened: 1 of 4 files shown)


be-java-springboot/files/README.md
```diff
@@ -0,0 +1,59 @@
...
+  stage('Build and Unit Test') {
+    withEnv(["TAGVERSION=${context.tagversion}",...,"GRADLE_USER_HOME=/.gradle-cache",...]) {
+      ...
+    }
+  }
+}
+...
+```
+
+**NOTE**: Be aware that, as the Jenkins jobs are sharing the same PVC, only one Jenkins job can run at a time.
```
braisvq1996, line 59: If the PVC is crated with ReadWriteMany access mode then it would not have this restriction right?

be-java-springboot/files/README.md
```diff
@@ -0,0 +1,59 @@
+# {{project-name}} - provisioned from the OpenDevStack Java/Spring Boot QuickStarter (be-java-springboot)
```
braisvq1996, line 1: If we want to have the name dynamically created, we need to process this file in the provisioning and use @component_id@ and @project_id@ instead

renedupont: Okay I see, I copied it from here: https://github.com/opendevstack/ods-quickstarters/blob/master/be-rust-axum/rust-template/README.md#project-name---from-the-opendevstack-rust-quickstarter-be-rust-axum
Not sure why it works there 😆 .
How do I process this file in the provisioning?

be-java-springboot/files/README.md
renedupont: Looking at the doc at https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes you seem to be right. Does it maybe also depend on the underlying storage provider type whether this is supported? Somehow I thought to remember that I had an issue with this once. But in case that works, that would be even better :)

stitakis: ReadWriteMany will allow different gradle processes to access in parallel the same cache folder. I'm not sure if this will have an negativ impact on the internal structure of cache. @renedupont do you know this?

renedupont: @stitakis I researched a bit and found this https://github.com/gradle/gradle/issues/851

Basically, there is no suitable solution for us with gradle to have multiple processes work on the same cache. Therefore I'd propose to leave the access mode and not change it to ReadWriteMany. Yes, this is a painful limitation but it is stated in the readme and I see this caching option more as an offer to everyone if they can cope with this limitation.

braisvq1996: It will for sure depend on the storage provider if it supports it. on our case NFS storage allow it but EBS do not.
Good point @stitakis, I am not sure if there would be any issue with RWX storage.....

be-java-springboot/files/README.md
braisvq1996: You can check this as an example:

- https://github.com/opendevstack/ods-quickstarters/blob/master/e2e-etl-python/Jenkinsfile#L37-L41
- https://github.com/opendevstack/ods-quickstarters/blob/master/e2e-etl-python/prod.yml.template

be-java-springboot/files/README.md
renedupont: The issue is not only the storage provider, but on top the locking mechanism of Gradle

be-java-springboot/files/README.md
renedupont: The doc (https://www.opendevstack.org/ods-documentation/opendevstack/latest/jenkins-shared-library/quickstarter-pipeline.html#_odsquickstarterstagerenderjenkinsfile) mentions specifically that this stage is for rendering the Jenkinsfile, I'm trying to render it in a README.md file though. I guess that would be quite a misuse of that stage in case it works at all, right? Maybe it is simpler just to replace it with something like `<project_id>` and don't render it?

braisvq1996: We already make use of that stage to render other files that are not the Jenkinsfile though :sweat_smile:

renedupont: Ok even though I'd suggest to rename this Rendering Stage then to make that clear, I used it now for the readme :)

be-java-springboot/files/README.md
renedupont: I extended the description of this limitation in the readme and moved it more to the top to make it more prominent.

PR #1055, opened 2024-09-04
Update OS packages by default in all Quickstarters

### LINKED ISSUE #1049: OS packages updated by default in the Quickstarters
Currently, when a Quickstarter gets provisioned, it comes with the base image, the specific language/tool for it and the example app. No other package is installed or updated and this could lead to some vuls reaching the end users

OS packages should be updated when the quickstarter gets provisioned for the 1º time by adding, for example, "yum update -y" on the dockerfile

N/A

**Additional context**
For the images that do not come with a package manager left them as they are

braisvq1996:
Closes #1049

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Update OS packages by default and bump gitleaks version ([#1049](https://github.com/opendevstack/ods-quickstarters/issues/1049))
 - Update Angular, Ionic and Typescript Quickstarters ([#1033](https://github.com/opendevstack/ods-quickstarters/issues/1033))
 - Update Rust Axum Quickstarter to Rust 1.79.0 ([#1024](https://github.com/opendevstack/ods-quickstarters/pull/1024))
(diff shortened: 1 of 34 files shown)


tbugfinder (approved):
LGTM

gerardcl (approved):
LGTM!
Regarding the UBI9 micro base image, it makes sense to not even try updating packages as IIRC there are almost no packages to worry about! Nevertheless, now and then, when building I scan with both trivy and aqua scanners and there are no findings, which make sense too. :+1:   (and, if I ever find something I will report ASAP)

braisvq1996:
> LGTM!
> Regarding the UBI9 micro base image, it makes sense to not even try updating packages as IIRC there are almost no packages to worry about! Nevertheless, now and then, when building I scan with both trivy and aqua scanners and there are no findings, which make sense too. :+1:   (and, if I ever find something I will report ASAP)

Yeah, I took a look at the amount of packages it has and it is only about 20. The other UBI9 images we use in contrast have over 100. And even if there is any vul, Red Hat usually patch it and pushes a new image that will be pulled next time any end user with the rust QS build its application

PR #1070, opened 2024-10-21
Cypress login with MFA

roicarrera:
New function for login into the application under test bypassing MFA by using otplib to retrieve a one-time password automatically
Removed obsolete msalv2 login function
Extracted login function out of e2e.ts


BraisVQ (approved):

metmajer:
@roicarrera @cschweikert @BraisVQ secrets with id/username and password should never be consumed via "oc get secret", which is unsafe, but via Jenkins' ``withCredentials`` to avoid credentials leakage into log files and leaving traces elsewhere. Please note for next time. I have asked this to be cleaned in https://github.com/opendevstack/ods-quickstarters/pull/1079.

PR #1079, opened 2024-11-06
Generate PDF report for cypress + environment management improvements

roicarrera:
added mochawesome
added pdf-generator script file
read baseurl on config maps based on environment
added skip for prod


cschweikert:
Hi @roicarrera, since this PR is also reworking parts of the respective configs, I would propose to use `cypress.config.ts` as base configuration that is then mixed into the other configs (import the base config and using the spread operator). This would prevent/remove a lot of code duplication and put more focus on the actual difference between those files.

e2e-cypress/files/package.json
```diff
@@ -21,12 +21,15 @@
...
     "cypress": "^13.13.1",
+    "cypress-multi-reporters": "^1.6.4",
     "junit-report-merger": "^7.0.0",
     "mocha-junit-reporter": "^2.2.1",
     "npm-run-all": "^4.1.5",
     "otplib": "^12.0.1",
+    "puppeteer": "^23.5.1",
     "rimraf": "^6.0.1",
     "sharp": "^0.33.5",
+    "ts-node": "^10.9.2",
```
cschweikert, line 32: I saw that `ts-node` is used with `npx` inside the `Jenkinsfile.template`. Either it should be removed here or maybe there should be a dedicated PDF task as part of `package.json` that is then called by `npm` instead of `npx`.

e2e-cypress/files/pdf-generator.ts
```diff
@@ -0,0 +1,68 @@
...
+      (testHeader as HTMLElement).click();
+    }
+  });
+}
+
+const isLocal = process.env.NODE_ENV === 'local';
+
+(async () => {
+  try {
+    const files = await fs.promises.readdir('build/test-results/mochawesome/');
```
cschweikert, line 18: The foldername 'build/test-results/mochawesome/' is used multiple times. Maybe this should be replaced by a constant or another path variable.

e2e-cypress/files/pdf-generator.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    for (const file of files) {
+
+      if (!fs.existsSync(path.resolve(__dirname, 'build/test-results/mochawesome/', 'pdf')))
+        fs.mkdirSync(path.resolve(__dirname, 'build/test-results/mochawesome/', 'pdf'));
```
cschweikert, line 23: Creating a new folder should move to the spot where it is first needed. This would prevent situations where a new folder is created without further content, e.g. when the loop ends earlier due to the `continue` in the next lines or a runtime error.

e2e-cypress/files/support/test-evidence.ts
```diff
@@ -65,6 +65,10 @@ export const takeScreenshotEvidence = (testName: string, testStep: number, testS
         logEvidence(testName, testStep, description, [
           `Stored screenshot "${path.basename(result.path)}" with hash (sha256) ${result.hash} taken at ${String(data.takenAt)} as evidence.`
         ]);
+
+        // Create a relative path from the screenshots folder to the mochawesome test-results folder
```
cschweikert, line 69: What exactly is happening here? A comment for that would be good. Also, is this a recursion with the risk of having an endless loop?

e2e-cypress/Jenkinsfile.template
```diff
@@ -79,16 +95,20 @@ def stageTest(def context) {
...
         stash(name: "acceptance-test-videos-${context.componentId}-${context.buildNumber}", includes: 'cypress/videos.zip', allowEmpty: true)
         archiveArtifacts artifacts: 'cypress/videos.zip', fingerprint: true, daysToKeep: 2, numToKeep: 3
       }
 
-      if (fileExists('cypress/screenshots')) {
-        zip zipFile: 'cypress/screenshots.zip', archive: false, dir: 'cypress/screenshots'
-        stash(name: "acceptance-test-screenshots-${context.componentId}-${context.buildNumber}", includes: 'cypress/screenshots.zip', allowEmpty: true)
-        archiveArtifacts artifacts: 'cypress/screenshots.zip', fingerprint: true, daysToKeep: 2, numToKeep: 3
+      if (fileExists('build/test-results/screenshots')) {
+        zip zipFile: 'screenshots.zip', archive: false, dir: 'build/test-results/screenshots'
```
cschweikert, line 109: Maybe this should be `cypress/screenshots.zip` to stay in sync with the other artifact creation (`cypress/pdf.zip`, `cypress/videos.zip`).

cschweikert (changes requested):

e2e-cypress/files/support/test-evidence.ts
roicarrera: The addScreenshot function is for adding an screenshot as context to the html reporter (it is defined in e2e.ts), it's not the same function as the takeScreenshotEvidence function, it's called at the end of it so that all screenshots taken in this way appear in the html report that's later converted to a pdf. Is this what you meant?

e2e-cypress/files/support/e2e.ts
```diff
@@ -15,3 +16,13 @@ afterEach(function() {
 
   consoleLogs.splice(0);
 })
+
+Cypress.Commands.add('addScreenshot', (title: string, screenshot: string) => {
```
cschweikert, line 20: I'd rename anything "screenshot" here to "context path", e.g., `addScreenshot`->`addContextPath`.

e2e-cypress/files/support/test-evidence.ts
cschweikert: Ah, now I got it. Thanks for clarifying :)

e2e-cypress/files/cypress-acceptance.config.ts
```diff
@@ -2,18 +2,30 @@ import { defineConfig } from 'cypress'
...
+      json: true
+    },
+    reportersCustomReporterJsReporterOptions: {
+      mochaFile: 'build/test-results/acceptance-junit-[hash].xml',
+      toConsole: true,
+    },
   },
   e2e: {
-    baseUrl: 'https://www.w3schools.com',
+    baseUrl: process.env.CYPRESS_BASE_URL,
```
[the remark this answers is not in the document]
roicarrera: You are right. I'll change the env variable name to BASE_URL so that it makes more sense. I prefer to keep the baseUrl parameter there as I think it's clearer for local env

e2e-cypress/files/pdf-generator.ts
roicarrera: True, I moved it in my last push. I think I addressed all your comments with it. Thanks for all of them!

e2e-cypress/files/support/test-evidence.ts
```diff
@@ -65,6 +65,10 @@ export const takeScreenshotEvidence = (testName: string, testStep: number, testS
...
+        const relativePath = path.relative('build/test-results/mochawesome', result.path.replace(/^.*(build.*)$/, '$1'));
```
cschweikert, line 70: also here I would replace the path with the variable from the configuration:
```suggestion
        const relativePath = path.relative(baseConfig.reporterOptions.mochawesomeReporterOptions.reportDir, result.path.replace(/^.*(build.*)$/, '$1'));
```

cschweikert (commented):
Thanks for making all the changes. Really improved things around the configs! Only little things left.

cschweikert (dismissed):
🚀

e2e-cypress/Jenkinsfile.template
```diff
@@ -29,6 +29,17 @@ odsComponentPipeline(
     // 'release/': 'test'
   ]
 ) { context ->
+
+  // Note: Testing in the production environment is not recommended as it can lead to unintended consequences,
+  // including potential downtime, data corruption, or exposure of sensitive information.
+  // This block is designed to skip tests in the production environment to avoid these risks.
+  // If you choose to enable testing in production, do so at your own risk and take all necessary precautions.
+  if (context.environment == 'prod') {
```
metmajer, line 37: @roicarrera the Cypress Quickstarter is of type ods-test, which must support acceptance, integration, and installation tests. You cannot simply omit all tests in production. Since installation tests are mandatory by LeVA in the PROD environment, you are right now messing with compliance regulations. Changing the current testing behavior in PROD needs a clear design in agreemtn with the core platform team and consistency across all ods-test Quickstarters. FYI @BraisVQ

e2e-cypress/Jenkinsfile.template
```diff
@@ -49,27 +60,32 @@ odsComponentPipeline(
 
 def stageTest(def context) {
   stage('Integration Test') {
-    // OPTIONAL: load environment variables for Azure SSO with MSALv2; please adapt variable names to your OpenShift config
+    // Define your DEV and QA base URLs in a config map in OpenShift; please adapt variable names to your OpenShift config
```
metmajer, line 63: @roicarrera why define the DEV and QA base urls in a secret if 12-factor apps teaches us to provide them through the environment? I believe that ODS already does that. Then, a test case should only reference a path and resouce below the BASE URL. Can you clarify the design? What about PROD?

e2e-cypress/Jenkinsfile.template
```diff
@@ -79,14 +95,18 @@ def stageTest(def context) {
...
         stash(name: "acceptance-test-videos-${context.componentId}-${context.buildNumber}", includes: 'cypress/videos.zip', allowEmpty: true)
         archiveArtifacts artifacts: 'cypress/videos.zip', fingerprint: true, daysToKeep: 2, numToKeep: 3
       }
 
-      if (fileExists('cypress/screenshots')) {
-        zip zipFile: 'cypress/screenshots.zip', archive: false, dir: 'cypress/screenshots'
+      if (fileExists('build/test-results/screenshots')) {
+        zip zipFile: 'cypress/screenshots.zip', archive: false, dir: 'build/test-results/screenshots'
         stash(name: "acceptance-test-screenshots-${context.componentId}-${context.buildNumber}", includes: 'cypress/screenshots.zip', allowEmpty: true)
         archiveArtifacts artifacts: 'cypress/screenshots.zip', fingerprint: true, daysToKeep: 2, numToKeep: 3
```
metmajer, line 111: @roicarrera what's the intended purpose of mochawesome reports opposed to LeVA-based documentation? Why archive reports, videos, and screenshots in Jenkins where they keep consuming disk space, even only for two days, which could cause Jenkins to fail when the disk runs full?

metmajer (changes requested):
Hello @roicarrera,

- We need a clear concept for generating reports, screenshots, and videos from our testing Quickstarters. I cannot see clearly how their creation and storing does not affect Jenkins. The only place to store potentially larger files is Nexus.
- Disabling test execution in PROD in the way you did seriously messes with regulations. A solid design involving the core platform team needs to be done to prevent potential damage.

As a general rule, I also suggest creating one PR for one purpose, and not mixing many different things. This will help increase clarity during code reviews.

e2e-cypress/Jenkinsfile.template
```diff
@@ -49,27 +60,32 @@ odsComponentPipeline(
...
+      // test: sh(returnStdout: true, script:"oc get configmaps cypress-config -o jsonpath='{.data.TEST_BASE_URL}'").trim()
+    ]
+
+    def baseUrl = baseUrls.get(context.environment ?: 'dev', 'https://www.w3schools.com') // default to W3Schools for demo purposes, replace with your own default
+
+    // Example for loading environment variables for Azure SSO; please adapt variable names to your OpenShift config,
+    // making sure to precede the variable names with the environment name in lowercase (e.g., dev_username, dev_password,
+    // test_username, test_password, etc.)
+    // cypressUser = sh(returnStdout: true, script:"oc get secret e2euser -o jsonpath='{.data.${context.environment}_username}' | base64 -d")
+    // cypressPassword = sh(returnStdout: true, script:"oc get secret e2euser -o jsonpath='{.data.${context.environment}_password}' | base64 -d")
```
metmajer, line 77: @roicarrera credentials (secrets of type username and password) need to be consumed in a safe way so that they do not leak into the log stream. Please use Jenkins' ``withCredentials`` task. You can refer to other Quickstarters to see its use.

metmajer (changes requested):

e2e-cypress/Jenkinsfile.template
metmajer: @roicarrera your comment says that testing in production means to take a risk and requires precaution. What do you mean? We made this conciously to ensure that applications, once in production, work as expected. Limiting test execution to installation tests is a possible and helpful improvement for the future for sure.

cschweikert:
>     * We need a clear concept for generating reports, screenshots, and videos from our testing Quickstarters. I cannot see clearly how their creation and storing does not affect Jenkins. The only place to store potentially larger files is Nexus.

Build artifacts like screenshots and videos also fit the purpose of analyzing problems in case of broken tests. Yes, you should assure that e2e tests are successful BEFORE you commit. This could be done by running them on a locally running representation of your target system. However, the situation on the cluster is different. In rare cases you are even facing things like flaky tests that are hard to reproduce even on the cluster. In those situations as a developer you are thankful for every piece of information you could get.

In this context they fit where they are. However, we should discuss about the right amount of "daysToKeep". Also it would be great to understand, if storing screenshots and videos temporarily even creates a problem regarding storage sizes. Maybe Jenkins already takes care of this.

e2e-cypress/Jenkinsfile.template
```diff
@@ -44,53 +45,105 @@ odsComponentPipeline(
...
+        // string(credentialsId: authenticatorOTPSecret, variable: 'OTP_SECRET')
+      ]) {
+        sh 'npm install'
+
+        // Note: Testing in the production environment is not recommended as it can lead to unintended consequences,
+        // including potential downtime, data corruption, or exposure of sensitive information.
+        // This block is designed to skip acceptance and integration tests in the production environment to avoid these risks.
+        // If you choose to enable these tests in production, do so at your own risk and take all necessary precautions.
+        def status
+        if (context.environment == 'prod') {
```
cschweikert, line 110: This split between `npm run e2e` and `npm run e2e:prod` might require an update in the documentation here: https://github.com/opendevstack/ods-quickstarters/blob/master/docs/modules/quickstarters/pages/e2e-cypress.adoc?plain=1#L124-L137

e2e-cypress/files/package.json
```diff
@@ -14,19 +14,25 @@
     "junit-acceptance-report": "jrm build/test-results/acceptance-junit.xml 'build/test-results/acceptance-*.xml'",
     "delete-junit-results": "rimraf build/test-results",
     "e2e": "npm-run-all delete-junit-results cypress:run-installation cypress:run-integration cypress:run-acceptance",
+    "e2e:prod": "npm-run-all delete-junit-results cypress:run-installation",
     "combine:reports": "npm-run-all junit-installation-report junit-integration-report junit-acceptance-report",
-    "e2e:jenkins:record": "npm run delete-junit-results && npm run cypress:run-installation -- --record && npm run cypress:run-integration -- --record && npm run cypress:run-acceptance -- --record"
+    "e2e:jenkins:record": "npm run delete-junit-results && npm run cypress:run-installation -- --record && npm run cypress:run-integration -- --record && npm run cypress:run-acceptance -- --record",
+    "e2e:jenkins:report:prod": "npm run delete-junit-results && npm run cypress:run-installation -- --record",
```
cschweikert, line 20: I guess the intention here was to have a prod version of `e2e:jenkins:record`:
```suggestion
    "e2e:jenkins:record:prod": "npm run delete-junit-results && npm run cypress:run-installation -- --record",
```

Like for the `e2e:prod` the documentation (https://github.com/opendevstack/ods-quickstarters/blob/master/docs/modules/quickstarters/pages/e2e-cypress.adoc?plain=1#L124-L137) might require an update reflecting flavors for the different target environments.

e2e-cypress/Jenkinsfile.template
```diff
@@ -55,7 +45,26 @@ odsComponentPipeline(
       ]
     )
   }
-
+  if (fileExists('cypress/screenshots.zip')) {
```
cschweikert, line 48: Maybe I'm missing something, but right now it seems that videos will go to a zip and nowhere else from there. Maybe there should be a block like the following as well:
```
  if (fileExists('cypress/videos.zip')) {
    odsComponentStageUploadToNexus(context,
      [
        distributionFile: 'cypress/videos.zip',
        repository: 'leva-documentation',
        repositoryType: 'raw',
        targetDirectory: "${targetDirectory}"
      ]
    )
  }
```

However, I don't see the need for videos being part of any documentation, if there are things like test protocols and screenshots already available. I still believe that it might be a good option to keep videos as build artifacts with Jenkins. Maybe we bring back the block
```
      if (fileExists('cypress/videos')) {
        zip zipFile: 'cypress/videos.zip', archive: false, dir: 'cypress/videos'
        stash(name: "acceptance-test-videos-${context.componentId}-${context.buildNumber}", includes: 'cypress/videos.zip', allowEmpty: true)
        archiveArtifacts artifacts: 'cypress/videos.zip', fingerprint: true, daysToKeep: 2, numToKeep: 3
      }
```
and set the `numToKeep` to 1 or leave it commented out. And we could add a warning as code comment that storage usage should be monitored when dealing with videos. Or we add a check on the file size against an upper limit before archiving the artifact.

cschweikert (changes requested):

e2e-cypress/Jenkinsfile.template
roicarrera: you are right! Forgot about this part of the documentation, thanks!

cschweikert:
I continued thinking about and like to come back to my previous comment regarding the purpose of the videos:

> Build artifacts like screenshots and videos also fit the purpose of analyzing problems ...

I still don't see the need for videos being part of any documentation, if there are things like test protocols and screenshots that can be linked to the responsible source code that defines the test scenarios. I'd recommend to keep videos as build artifacts with Jenkins and rather restrict the `numToKeep` to 1. This would require Jenkins to store one video ZIP in total. With this the risk of any storage issues is quite low.

More important is the following: Imagine every little e2e test run in the cluster leads to an upload of a big video file into Nexus. This will cause additional load in the network, puts pressure on the Nexus instance and consumes a good amount of storage capacity. I'm not aware of how long the concrete documents will be stored in Nexus. If there is no limited data retention, this will cause ever increasing resource allocation, which I don't see to be a sustainable option.

@roicarrera @metmajer What are your thoughts about this?

metmajer:
@cschweikert

> Build artifacts like screenshots and videos also fit the purpose of analyzing problems in case of broken tests.

There is nothing wrong with providing test execution results, screenshots, and videos. As a former software engineer, I can relate to what you say. Since Jenkins is a SPoF, we need to align before using Jenkins as a file store, even for short-term storage. Currently, Jenkin's file system is only 50 MB in size and would exhaust after only a few pipeline runs. As a result, Jenkins would stop working and would refuse to start until disk space has been freed up. Also, our support team would be busy handling support tickets. Additionally, I want to ensure that additional test reports align with the SLC documents we already produce to ensure that the benefit of either type of report is understood internally, and, by our users.

> I still don't see the need for videos being part of any documentation

I agree. I believe that we should have a solid reasoning for videos. Whatever we put into our Quickstarters should appeal to 80% of our users, and not be an edge case. Capable users will be able to integrate support for videos should they need it.

FYI @roicarrera

metmajer:
I recently shared the following feedback with @roicarrera. Sharing it here for transparency and record. FYI @cschweikert 

Those are not only integration tests
https://github.com/opendevstack/ods-quickstarters/blob/3f51c8a30e835b67599b84cd9aa379f3f0725c4b/e2e-cypress/Jenkinsfile.template#L71

--
How do we make users understand what they have to do here? - Why not leave them commented in and simply educate that the config map must be populated / fail the pipeline if they are not set?
https://github.com/opendevstack/ods-quickstarters/blob/3f51c8a30e835b67599b84cd9aa379f3f0725c4b/e2e-cypress/Jenkinsfile.template#L74-L79

--
You need to use a credentials secret in OpenShift and use Jenkins' withCredentials to safely use the secret in transit.
https://github.com/opendevstack/ods-quickstarters/blob/3f51c8a30e835b67599b84cd9aa379f3f0725c4b/e2e-cypress/Jenkinsfile.template#L86-L87

--
These are a lot of commented lines; why not build better logic around them? Quickstarters should be quick to use, not require a deeper understanding of what's going on inside them.
https://github.com/opendevstack/ods-quickstarters/blob/3f51c8a30e835b67599b84cd9aa379f3f0725c4b/e2e-cypress/Jenkinsfile.template#L83-L88
and https://github.com/opendevstack/ods-quickstarters/blob/3f51c8a30e835b67599b84cd9aa379f3f0725c4b/e2e-cypress/Jenkinsfile.template#L99-L101

--
Stashes are for communicating data to the Release Manager. We should only use them if needed. For now, there is no contract defined for exchanging screenshots between e2e-test components and the RM. Let's do this only once the contract has been defined.
https://github.com/opendevstack/ods-quickstarters/blob/3f51c8a30e835b67599b84cd9aa379f3f0725c4b/e2e-cypress/Jenkinsfile.template#L143

--
We need to avoid storing files in Jenkins. Quickstarters can too quickly fill up Jenkins and Jenkins is only an executor, not a file store. For this we have Nexus.
https://github.com/opendevstack/ods-quickstarters/blob/3f51c8a30e835b67599b84cd9aa379f3f0725c4b/e2e-cypress/Jenkinsfile.template#L144

--
The return statement does not do anything. Please align with the behavior of Spock/Geb. A failed attempt to execute tests must fail the component. See https://github.com/opendevstack/ods-quickstarters/blob/master/e2e-spock-geb/Jenkinsfile.template#L32-L34 for reference.
https://github.com/opendevstack/ods-quickstarters/blob/3f51c8a30e835b67599b84cd9aa379f3f0725c4b/e2e-cypress/Jenkinsfile.template#L147

cschweikert:
> Currently, Jenkin's file system is only 50 MB in size and would exhaust after only a few pipeline runs.

Hm, I wonder if "50 MB" is a typo. When looking into the different *-cd OC projects I saw volume claims of 5 GB or 10 GB.

I did some quick experiments regarding filesize of those video files. A test file that does full page reloads every 2 seconds showing different websites in the resolution configured in this QS. Extrapolated to a 60 minutes test run the filesize would be about 0.5 GB. This is about the size of the `node_modules` folder of an average SPA frontend project.

> Capable users will be able to integrate support for videos should they need it.

I like the idea. I would recommend to remove any record-specifc scripts (e.g. `e2e:jenkins:record`). This would also make the documentation a bit simpler. At the same time, I would suggest to keep the following code:
```
if (fileExists('cypress/videos')) {
  zip zipFile: 'cypress/videos.zip', archive: false, dir: 'cypress/videos'
  archiveArtifacts artifacts: 'cypress/videos.zip', fingerprint: true, daysToKeep: 1, numToKeep: 1
}
```
It will only trigger when the user enables the recording feature. Having it in place is helpful and would also reflect our "opinion" regarding a good setting for things like `numToKeep`.

roicarrera:
> I like the idea. I would recommend to remove any record-specifc scripts (e.g. `e2e:jenkins:record`). This would also make the documentation a bit simpler. At the same time, I would suggest to keep the following code:

Regarding this, the --record flag, and subsequently the e2e:jenkins:record command are for using Cypress Cloud, not for recording videos. The video feature is enabled in the cypress config file. I agree with leaving the video disabled by default tho, and having the users enable it if wanted, but the record command is a separate feature that does not affect Jenkins space

metmajer:
@roicarrera @cschweikert I can see that the most critical feedback has been addressed. Points left for potential improvement are:
- commenting/uncommenting of lines; can you think of a better approach that transparently uses configmaps/secrets once defined by the user?
- `daysToKeep` and `numToKeep`; looking at https://github.com/jenkinsci/jenkins/blob/e2999799a9f983b302c63d7aaf7856f4da62e7d4/core/src/main/java/hudson/tasks/ArtifactArchiver.java#L391, I believe that the named properties are ignored completely? They are declared only in https://github.com/jenkinsci/jenkins/blob/e2999799a9f983b302c63d7aaf7856f4da62e7d4/core/src/main/java/hudson/tasks/LogRotator.java#L269, but class `ArtifactArchiver` neither inherits nor is composed of a `LogRotator`. Googling also didn't show any examples of `archiveArtifact` in combination with these properties. Thoughts?

e2e-cypress/files/README.md
```diff
@@ -10,38 +10,23 @@ Please note that each stage is executed with its own Cypress configuration file
...
 ## Reports
 From [Merging reports across spec files](https://docs.cypress.io/guides/tooling/reporters#Merging-reports-across-spec-files): each spec file is processed completely separately during each cypress run execution. Thus each spec run overwrites the previous report file. To preserve unique reports for each spec file, use the `[hash]` in the `mochaFile` filename.
 
 In order to generate one xml report per test type (installation, integration and acceptance) we use the junit-report-merger tool. See also the `junit-...-report` tasks in `package.json`.
 
 ## E2e test user authentication
 
-With Cypress 12 version is now available `cy.origin()` that allows you to handle redirections. This funcionality eases the login handling.
-See `./support/e2e.ts` for a generic login example.
+With Cypress 12 version is now available `cy.origin()` that allows you to handle redirections. This functionality eases the login handling.
```
cschweikert, line 26:
```suggestion
Starting version 12 of Cypress `cy.origin()` allows you to handle redirections. This functionality eases the login handling.
```

e2e-cypress/files/README.md
```diff
@@ -10,38 +10,23 @@ Please note that each stage is executed with its own Cypress configuration file
...
+See `./support/login-functions.ts` for a generic login example.
+
+In order to load your testing user's credentials for this login with SSO, you need to create a secret in OpenShift with the label `credential.sync.jenkins.openshift.io=true` and import them as environment variables using the withCredentials block to keep them secure. Find an example on how to do this in the `Jenkinsfile`.
```
cschweikert, line 29:
```suggestion
In order to load your testing user's credentials for this login with SSO, you need to create a secret in OpenShift with the label `credential.sync.jenkins.openshift.io=true` and import them as environment variables using the `withCredentials` block to keep them secure. Find an example on how to do this in the `Jenkinsfile`.
```

metmajer:
Well done @roicarrera and thanks @cschweikert for the support -- LGTM at last :)

metmajer (approved):

PR #1110, opened 2025-03-14
Fix type of no_nexus

### LINKED ISSUE #1109: be-spring-boot QS: no_nexus interpreted as string instead of boolean
**Describe the bug**
**be-spring-boot quickstarter**
In the generated `build.gradle`, the `no_nexus` param is a `String` instead of a `Boolean`. 
No matter what the string contains, evaluated as `Boolean` any value will result in `True`.
 
When it is then used to decide whether Nexus or mavenCentral is used as repo and is negated (in code`if (!no_nexus) {`), it will always result in `False` and thus the `else` block will always be executed.

**Expected behavior**
`no_nexus` should be a Boolean type.

**Additional context**
Will provide a PR soon...

fkressmann:
and add logs to show during runtime which repo is used (helps findig issues there faster)

Fixes #1109

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 
 ### Fixed
 
+- Fixed no_nexus interpreted as string instead of boolean for Java QS ([#1109](https://github.com/opendevstack/ods-quickstarters/issues/1109))
+
 ## [4.7.2] - 2025-2-03
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


metmajer (dismissed):

BraisVQ (approved):
tested and it works as it should now, going to nexus instead of Maven

PR #1112, opened 2025-03-19
e2e-cypress: get rid of additional chrome installation

matzehecht:
The e2e-cypress quickstarter currently includes Puppeteer as dependency. The Puppeteer library always downloads its own copy of chrome. This has the advantage that we can be sure that the Chrome version is compatible with Puppeteer. BUT we do not use that downloaded copy (as we specify the `executablePath` of the global installation of chrome: `/usr/bin/google-chrome`). Therefore, I see no reason why we should need that additional copy of chrome (which is installed by Puppeteer). 
By switching from `puppeteer` to `puppeteer-core` we skip the installation of chrome during the postinstall step. In fact, `puppeteer` is using `puppeteer-core` in the background. The only differences between those to packages is the former one installs chrome and it can be configured with some `PUPPETEER_*` env vars which are ignored by the latter one (and not used by us; References: https://devdocs.io/puppeteer/ and https://how.dev/answers/puppeteer-vs-puppeteer-core).

This change reduced the time needed for the install dependencies stage from ~ 1 min to ~ 18 seconds in my case (not a representative study 😉)

I also switched from chrome to edge for the pdf generation. Therefore, the whole e2e-cypress quickstarter only depends on one browser (I would be open to revert that change).

Closes #...
Fixes #...

Tasks: 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Set Request and Limits with Ephemeral storage and Make use of Nexus in terraform agents ([#1104](https://github.com/opendevstack/ods-quickstarters/pull/1104))
 - Fix ds-streamlit uses old CI colors ([#978](https://github.com/opendevstack/ods-quickstarters/issues/978))
 - Remove unnecessary evidence printing in e2e-spock-geb ([#1106](https://github.com/opendevstack/ods-quickstarters/pull/1106))
+- e2e-cypress get rid of additional chrome installation and switch to edge for pdf generation
 
 ### Fixed
 
--- a/e2e-cypress/files/package.json
+++ b/e2e-cypress/files/package.json
@@ -30,7 +30,7 @@
     "mochawesome": "^7.1.3",
     "npm-run-all": "^4.1.5",
     "otplib": "^12.0.1",
-    "puppeteer-core": "^23.5.1",
+    "puppeteer-core": "^24.4.0",
     "rimraf": "^6.0.1",
     "sharp": "^0.33.5",
     "ts-node": "^10.9.2",
--- a/e2e-cypress/files/package.json
+++ b/e2e-cypress/files/package.json
@@ -30,7 +30,7 @@
     "mochawesome": "^7.1.3",
     "npm-run-all": "^4.1.5",
     "otplib": "^12.0.1",
-    "puppeteer": "^23.5.1",
+    "puppeteer-core": "^23.5.1",
     "rimraf": "^6.0.1",
     "sharp": "^0.33.5",
     "ts-node": "^10.9.2",
--- a/e2e-cypress/files/pdf-generator.ts
+++ b/e2e-cypress/files/pdf-generator.ts
@@ -26,7 +26,7 @@ const isLocal = process.env.NODE_ENV === 'local';
         continue;
       }
 
-      const executablePath = isLocal ? undefined : '/usr/bin/google-chrome';
+      const executablePath = isLocal ? undefined : '/usr/bin/microsoft-edge';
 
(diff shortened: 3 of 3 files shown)


cschweikert (dismissed):
Thank you for getting rid of this overhead 🚀
Would be great to also change puppeteer-core to latest version `24.4.0`.

matzehecht:
> Thank you for getting rid of this overhead 🚀 Would be great to also change puppeteer-core to latest version `24.4.0`.

Done.

CHANGELOG.md
```diff
@@ -300,7 +319,7 @@
 - fix typescript-express junit test location ([#654](https://github.com/opendevstack/ods-quickstarters/issues/654))
 - fix java not in path for python quickstarter ([#685](https://github.com/opendevstack/ods-quickstarters/issues/685))
 - fix gitignore in inf-terraform ([#767](https://github.com/opendevstack/ods-quickstarters/issues/767))
-- fix e2e-spock-geb quickstarter groovy tests runs twice ([#874] https://github.com/opendevstack/ods-jenkins-shared-library/issues/874)
+- fix e2e-spock-geb quickstarter groovy tests runs twice ([#874] <https://github.com/opendevstack/ods-jenkins-shared-library/issues/874>)
```
cschweikert, line 322: It seems a formatter did some changes on this file. Here there actually is a problem with the link.
```suggestion
- fix e2e-spock-geb quickstarter groovy tests runs twice ([#874](https://github.com/opendevstack/ods-jenkins-shared-library/issues/874))
```
In other places there had been additional `<...>` around links as well. I'm not sure, if this is valid markdown.

cschweikert (changes requested):

CHANGELOG.md
matzehecht: Thanks. Yes, my formatter is configured differently. I just reverted the unnecessary changes.

cschweikert (dismissed):

garcanam (approved):

garcanam (approved):

metmajer:
@matzehecht @garcanam @BraisVQ do you assume that we can remove support for Chrome because you are not actively using it? How would users use Chrome if they see a need for it?

matzehecht:
Hey @metmajer, I am not sure I understand your question correctly. Therefore I will answer extensively (to answer the multiple ways I understood the questions).
What this PR did was switching from `puppeteer` to `puppeteer-core` as the former one installs its own chrome installation as soon as you run `npm install`. This installation was never used as chrome (and ms-edge) is installed in the jenkins agent. cypress and even the pdf-generator were always configured to use the central installation of chrome. The second thing done in this PR is switching the pdf-generator to ms-edge, to be using the same browser for running the tests and generating the report. With that this quickstarter is not using chrome anymore. 

So, if the question is if/why we removed chrome in this PR. We only removed a secondary installation. The original one is still implemented in the jenkins agents (e.g. [here](https://github.com/opendevstack/ods-quickstarters/blob/8c73e25bfbd2679af54eb17da9914bbcfa118324/common/jenkins-agents/nodejs22/docker/Dockerfile.ubi8#L46)). Chrome can still be found in its default installation location (`/usr/bin/google-chrome`).

If the question is if I still see a need to install chrome in the jenkins-agent: I am not the right one to answer this question, as I do not know if it is used somewhere else. For users of the e2e-cypress quickstarter I personally see no reason to use chrome as ms-edge is included (and is also a chromeium based browser).

metmajer:
Hey @matzehecht, and thanks for the clarification! I had misinterpreted the PR as a likely removal of an otherwise useful option.

PR #1114, opened 2025-03-19
e2e-cypress small improvements

matzehecht:
This PR includes the following improvements for e2e-cypress (as suggestion; I am happy about feedback)
- it switches from `npm install` to `npm clean-install` (`npm ci`)
  `npm install` is usually not a good idea in a continuous integration pipeline as it may update packages. `npm clean-install` will ensure it does not update any packages and will only use what is defined in the package-lock.json (means the exact version used by the developer).
- the current released version has some TypeScript errors (without any change to the TypeScript settings). Those are fixed by enabling `skipLibCheck` and fixing one type ([here](https://github.com/opendevstack/ods-quickstarters/commit/60a5fe78c60b01b2eaf89a455320a34847130acb#diff-47328c3ccece2c1955b36db3edd3a2141c7f84426d2dccdd42b32dc4ab29ee48R81))
- enables the strict rule set (as recommended)
  It also fixes all findings introduced by that
- Adds type checking to Jenkins to support reviews (by ensuring the type check succeeds)

Closes #...
Fixes #...

Tasks: 

### DIFF
--- a/e2e-cypress/files/support/e2e.ts
+++ b/e2e-cypress/files/support/e2e.ts
@@ -8,7 +8,7 @@ beforeEach(function() {
 })
 
 afterEach(function() {
-  const testName = this.currentTest?.fullTitle().replace(/ /g, '_');
+  const testName = this.currentTest?.fullTitle().replace(/ /g, '_') || 'unknown';
   const fileName = `system-output-${String(testName)}.txt`;
   const filePath = `cypress/results/${fileName}`;
 
--- a/e2e-cypress/files/support/e2e.ts
+++ b/e2e-cypress/files/support/e2e.ts
@@ -8,8 +8,8 @@ beforeEach(function() {
 })
 
 afterEach(function() {
-  const testName = this.currentTest.fullTitle().replace(/ /g, '_');
-  const fileName = `system-output-${testName}.txt`;
+  const testName = this.currentTest?.fullTitle().replace(/ /g, '_');
+  const fileName = `system-output-${String(testName)}.txt`;
   const filePath = `cypress/results/${fileName}`;
 
   cy.writeFile(filePath, consoleLogs.join('\n'));
--- a/e2e-cypress/files/support/e2e.ts
+++ b/e2e-cypress/files/support/e2e.ts
@@ -9,7 +9,7 @@ beforeEach(function() {
 
 afterEach(function() {
   const testName = this.currentTest?.fullTitle().replace(/ /g, '_') || 'unknown';
-  const fileName = `system-output-${String(testName)}.txt`;
+  const fileName = `system-output-${testName}.txt`;
   const filePath = `cypress/results/${fileName}`;
 
   cy.writeFile(filePath, consoleLogs.join('\n'));
--- a/e2e-cypress/files/support/login-functions.ts
+++ b/e2e-cypress/files/support/login-functions.ts
@@ -78,8 +78,18 @@ export function addLoginToAAD() {
 export function addSessionLoginWithMFA() {
   Cypress.Commands.add('sessionLoginWithMFA', (username: string, password: string) => {
     cy.session('login', () => cy.loginToAADWithMFA(username, password), {
-      validate: () => cy.getAllLocalStorage().should(validateLocalStorage),
-      cacheAcrossSpecs: true
+      validate: () => new Promise((resolve, reject) => {
+        try {
+          cy.getAllLocalStorage()
+            .should(validateLocalStorage)
+            .then(() => {
+              resolve()
+            })
+        } catch (e) {
+          reject(e)
+        }
+      }),
+      cacheAcrossSpecs: true,
     })
   })
 }
--- a/e2e-cypress/files/support/login-functions.ts
+++ b/e2e-cypress/files/support/login-functions.ts
@@ -98,8 +98,7 @@ export function addLoginToAADWithMFA() {
   Cypress.Commands.add('loginToAADWithMFA', (username: string, password: string) => {
     loginViaAAD(username, password);
     cy.getTOTP().then((otp) => {
-      const objOTP = { otp }
-      cy.origin('https://login.microsoftonline.com/', { args: objOTP }, ({ otp }) => {
+      cy.origin('https://login.microsoftonline.com/', { args: { otp } }, ({ otp }) => {
         cy.get("[name='otc']").type(otp)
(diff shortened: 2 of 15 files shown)


e2e-cypress/files/support/e2e.ts
```diff
@@ -8,8 +8,8 @@ beforeEach(function() {
 })
 
 afterEach(function() {
-  const testName = this.currentTest.fullTitle().replace(/ /g, '_');
-  const fileName = `system-output-${testName}.txt`;
+  const testName = this.currentTest?.fullTitle().replace(/ /g, '_');
```
cschweikert, line 11: I'd rather default that to an explicit string instead of having the `String(testName)` one lines below.
```suggestion
  const testName = this.currentTest?.fullTitle().replace(/ /g, '_') && 'unknown';
```
Mind the `&&` instead of the `??`. This should handle cases where the test name would be an empty string.

e2e-cypress/files/support/login-functions.ts
```diff
@@ -78,17 +78,27 @@ export function addLoginToAAD() {
...
 }
 
 export function addLoginToAADWithMFA() {
   Cypress.Commands.add('loginToAADWithMFA', (username: string, password: string) => {
     loginViaAAD(username, password);
-    cy.getTOTP().then((otp1) => {
-      const objOTP = { otp: otp1 }
+    cy.getTOTP().then((otp) => {
+      const objOTP = { otp }
       cy.origin('https://login.microsoftonline.com/', { args: objOTP }, ({ otp }) => {
```
cschweikert, line 102: I would not use the additional const here and do `{ args: { otp }}` directly:
```suggestion
      cy.origin('https://login.microsoftonline.com/', { args: { otp } }, ({ otp }) => {
```

e2e-cypress/files/tests/acceptance/acceptance.spec.cy.ts
```diff
@@ -20,12 +20,12 @@ describe('W3 application test', () => {
   it('Application is reachable', function () {
     cy.visit('/html/tryit.asp?filename=tryhtml_basic_paragraphs');
     cy.title().should('include', 'Tryit Editor');
-    printTestDOMEvidence(this.test.fullTitle(), 1, '#textareaCode', 'code area');
-    printTestDOMEvidence(this.test.fullTitle(), 2, '#iframecontainer', 'rendered code area');
-    takeScreenshotEvidence(this.test.fullTitle(), 3, 1, 'screenshot');
-    takeScreenshotEvidence(this.test.fullTitle(), 3, 2, 'screenshot substep 2');
+    printTestDOMEvidence(this.test?.fullTitle() ?? '', 1, '#textareaCode', 'code area');
```
cschweikert, line 23: I'd use a const for `this.test?.fullTitle()`. I'd also set the const with a default value, in case it is empty/undefined:
```
const testName = this.test?.fullTitle() && 'unknown';
```
also here: Mind the `&&`.

An alternative could be to skip the whole evidence creation, if the title is not given, or even throw an error.

e2e-cypress/files/support/login-functions.ts
```diff
@@ -78,17 +78,27 @@ export function addLoginToAAD() {
 export function addSessionLoginWithMFA() {
   Cypress.Commands.add('sessionLoginWithMFA', (username: string, password: string) => {
     cy.session('login', () => cy.loginToAADWithMFA(username, password), {
-      validate: () => cy.getAllLocalStorage().should(validateLocalStorage),
-      cacheAcrossSpecs: true
+      validate: () => new Promise((resolve, reject) => {
```
cschweikert, line 81: Would that work with async-await? Maybe like this:
```
validate: async () => {
  await cy.getAllLocalStorage().should(validateLocalStorage);
},
```

cschweikert (changes requested):

e2e-cypress/files/support/login-functions.ts
matzehecht: This does not work out of 2 reasons:

1. The validate function has to have a return type `Promise<false | void> | void` but the should function does not return a boolean (and not a boolean wrapped in a promise)
2. The async implementation of cypress is not based on native promises but on Chainables and cypress promises which have a then function but is not compatible with native promises (and async/await)

e2e-cypress/files/support/e2e.ts
matzehecht: Just one correction: We have to use `||` to do replace falsy values.

e2e-cypress/files/support/e2e.ts
```diff
@@ -8,8 +8,8 @@ beforeEach(function() {
 })
 
 afterEach(function() {
-  const testName = this.currentTest.fullTitle().replace(/ /g, '_');
-  const fileName = `system-output-${testName}.txt`;
+  const testName = this.currentTest?.fullTitle().replace(/ /g, '_') || 'unknown';
+  const fileName = `system-output-${String(testName)}.txt`;
```
cschweikert, line 12: I guess the wrapping `String()` is not needed anymore and you can revert the change to:
```suggestion
  const fileName = `system-output-${testName}.txt`;
```

e2e-cypress/files/support/test-evidence.ts
```diff
@@ -17,7 +17,7 @@ const logEvidence = (name: string, step: number, description: string, evidenceLo
   });
 }
 
-export const printTestDOMEvidence = (testName: string, testStep: number, selector: string, description: string) => {
+export const printTestDOMEvidence = (testName = 'unknown', testStep: number, selector: string, description: string) => {
```
cschweikert, line 20: TBH I would not introduce default values at this spot. It has the following disadvantages:
* `'unknown'` is a more or less opinionated term that would now be active for all projects using this.
* The input `testName` now has type `string | undefined`. This allows accidentially handing in a potentially undefined value. Using a default value at this spot would make it even harder to spot the mistake.
* This function and also the others have the purpose to provide per-test evidence. Offering the possibility to miss out `testName` doesn't make sense for this purpose.

I would leave `testName` without default setting, but use `'unknown'` when handing in the test name in `acceptance.spec.cy.ts`. There I would also make use of a const.

cschweikert (dismissed):

roicarrera (approved):

garcanam (dismissed):

cschweikert (approved):

garcanam (approved):

PR #1117, opened 2025-03-28
Dockerhub removal

BraisVQ:
Move away from DockerHub due to changes in pull rate limits: https://docs.docker.com/docker-hub/usage/
UPDATE: they reverted the new policy but eitherway we should move away from Dockerhub if possible: https://github.com/docker/docs/pull/22326

Switch to UBI or AWS ECR

Gateway/nginx Quickstarter alternative image for openrestry does not work properly so it ramains with Dockerhub image for now

### DIFF
--- a/be-fe-mono-repo-plain/files/docker_be/Dockerfile
+++ b/be-fe-mono-repo-plain/files/docker_be/Dockerfile
@@ -1,16 +1,8 @@
 # add /overwrite FROM with your base image, and do whatever you like here :)
-FROM registry.redhat.io/ubi9-minimal
+FROM registry.redhat.io/ubi9-micro
 
 RUN echo "building simple backend container"
 
-USER root
- 
-RUN microdnf -y update && \
-	microdnf install -y nc && \
-    microdnf clean all
- 
-USER 1001
-
 EXPOSE 8081
 
-CMD ["/bin/sh", "-c", "/usr/bin/nc -lk -p 8081 --sh-exec \"echo -e 'HTTP/1.1 200 OK\n\nHello World!\n$(date)'\""]
+CMD ["sh", "-c", "while true; do echo 'Container running on port 8081'; sleep 5; done"]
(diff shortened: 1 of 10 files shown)


be-golang-plain/files/docker/Dockerfile
```diff
@@ -1,33 +1,7 @@
-FROM alpine
-
-# Uncomment this section if you need to use glibc in your container, f.e. when you enable CGO_ENABLED=1
```
henrjk, line 3: Did you already check if glibc is in the base image as otherwise this would cause  https://github.com/opendevstack/ods-quickstarters/issues/277 ?

henrjk (commented):
Hi Brais, I hope my comments help move this along.?

be-fe-mono-repo-plain/files/docker_be/Dockerfile
```diff
@@ -1,12 +1,8 @@
...
 RUN echo "building simple backend container"
 
-RUN apk update  && \
-    apk -i upgrade && \
-    apk cache clean
-
 EXPOSE 8081
 
-CMD ["/bin/sh", "-c", "/usr/bin/nc -lk -p 8081 -e echo -e \"HTTP/1.1 200 OK\n\nHello World!\n$(date)\""]
+CMD ["sh", "-c", "while true; do echo 'Container running on port 8081'; sleep 5; done"]
```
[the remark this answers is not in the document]
BraisVQ: We can switch to minimal for mono repo and docker plain yes, It was also on of my doubt.
Should we also make the change for golang ?

henrjk (dismissed):

be-golang-plain/files/docker/Dockerfile
BraisVQ: Glibc package is not avaialble in a more up to date version in the repositories to install it for what I could see.
Building from source is also complicated (not successful in my tests) and resource-hungry process.

For now this glibc version should be enough for use cases

henrjk (approved):
LGTM

PR #1120, opened 2025-05-02
Feat - Rust rustup dynamic toolchain selection enablement

### LINKED ISSUE #1107: Rust QS - cargo deny issues
**Describe the bug**
Cargo deny in the pipeline does not work properly in certain regions due to network restrictions 

**To Reproduce**
1. Go to provision rust QS in a region with network restrictions
2. Check component pipeline execution
3. See error

**Expected behavior**
cargo deny should work out of the box

**Additional context**
We will need to make the stage optional so users can take advantage of it and the component works properly in all regions

### LINKED ISSUE #1119: Rust QS enable dynamic toolchain/version support
We currently fix the version supported of the Rust compiler and teams should be able to define their required version as per needs. Enabling backward and forward compatibility of the Jenkins CICD using Rust, with existing and new projects.

Since [`rustup`](https://www.rust-lang.org/tools/install) is the recommended tool to manage Rust toolchains/installations ODS should also provide it, hence enabling dynamic load of required projects' targets. 

Example:
1. New ODS Rust Jenkins Agent by default would provide support for a specific Rust toolchain/version and build tools.
2. A team with existing Rust components would be forced to update their Rust version, with the current approach.
3. With the new approach, by having `rust-toolchain` file in the root of their repo/component defining which version they use then that would make Jenkins (and any other team member developing on that component locally) dynamically load and use the right Rust version (thanks to `rustup`) when using the build tooling (`cargo`).

N/A

**Additional context**
In any case, we should provide a default official supported version.

gerardcl:
Description here ...

Closes #1119 
Fixes #1107 

Tasks: 


ManuelFeller (approved):

ManuelFeller (approved):

ManuelFeller (approved):

docs/modules/quickstarters/pages/be-rust-axum.adoc
```diff
@@ -190,11 +193,11 @@ The Jenkinsfile is provisioned with this quick starter to ease CI/CD process. In
 
 include::partials$secret-scanning-with-gitleaks.adoc
 
-== Builder agent used
+== Builder agent used by the Jenkins CICD
```
tbugfinder, line 196: typically we use CI/CD, no?

gerardcl: yes, that is implicit as stated in all the other quickstarters, if we want to change it here we shall change it everwhere, I close this as I don't see it required and nevertheless could be tackled in another PR later, ok?

PR #1121, opened 2025-05-12
Fix in the Python/Flask Quickstarter Unit tests

BraisVQ:
This is a fix for the situation when the Unit tests has any failure, to ensure that the result file is moved to the proper place regardless of the test result


gerardcl (approved):
LGTM! 👍

PR #1138, opened 2025-11-21
Fix parameter name for Spring Boot version and update to 4.0.0

BraisVQ:
According to the documentation, the parameter used when using springboot starter needs to be "bootVersion" and not "platformVersion"

Right now, as it is not valid, it is always provisioning the latest springboot version as the latest is 4.0.0 and requires Gradle 8.14 or above. The Gradle build in the component pipeline is broken at this moment for newly provisioned Quickstarters

Source: https://github.com/spring-io/start.spring.io/blob/main/USING.adoc#command-line-support


BraisVQ:
We not only have an issue setting the version for Springboot version, but the version we had is no longer supported in the srping start app. We need to update it

PR #1141, opened 2025-12-27
Provide rhel9 Dockerfile for agents

BraisVQ:
Depends on https://github.com/opendevstack/ods-core/pull/1336

TEST:

- [x] All agents successfully built with rhel9 agent base
- [x] Test Rust agent
- [x] Test Python agent
- [x] Test Node agents
- [x] Test Terraform agents
- [x] Test Scala agent
- [x] Test Jdk agent
- [x] test Golang agent

We cannot build the agents in GH actions as there is no base rhel9 agent (there is no public rhel9 agent to build our agent base)

### DIFF
--- a/common/jenkins-agents/python/docker/Dockerfile.ubi9
+++ b/common/jenkins-agents/python/docker/Dockerfile.ubi9
@@ -8,7 +8,6 @@ ARG nexusAuth
 ENV PYTHONUNBUFFERED=1 \
     PYTHONIOENCODING=UTF-8 \
     UV_INDEX_URL=https://${nexusAuth}@${nexusHost}/repository/pypi-all/simple \
-    UV_TRUSTED_HOST=${nexusHost} \
     PIP_TRUSTED_HOST=${nexusHost} \
     UV_CA_CERTIFICATE=/etc/ssl/certs/ca-bundle.crt \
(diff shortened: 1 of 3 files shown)


BraisVQ:
@terra-naut @tbugfinder could you check if the rhel9 terraform agents works properly ?

common/jenkins-agents/python/docker/Dockerfile.ubi9
```diff
@@ -0,0 +1,42 @@
...
+
+RUN yum install -y autoconf automake gcc gcc-c++ openssl-devel libffi-devel && \
+    curl -LsSf https://astral.sh/uv/install.sh | sh && \
+    mkdir -p $HOME/.config/pip && \
+    echo "[global]" > $HOME/.config/pip/pip.conf && \
+    echo "index-url = https://${nexusAuth}@${nexusHost}/repository/pypi-all/simple" >> $HOME/.config/pip/pip.conf && \
+    echo "trusted-host = ${nexusHost}" >> $HOME/.config/pip/pip.conf && \
+    yum -y clean all
+
+RUN for pyVersion in 3.8 3.9 3.11 3.12; do \
```
gerardcl: we should remove support for 3.8 and 3.9 already as they are EOL, maybe 3.9 we could keep one year max if really required. But, I would not offer those in EOL already but rather let users who really need add it themselves, or?

PR #1147, opened 2026-02-16
Increase default memory resources for Docker-plain Quickstarter

BraisVQ:
Increase Docker Plain default memory resources as by default container restarts and reaches OOMKilled

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Changed
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


gerardcl (approved):
LGTM! 🚀 
we shall now migrate the other quickstarter to helm, or? 💯

BraisVQ:
> LGTM! 🚀 we shall now migrate the other quickstarter to helm, or? 💯

Yeah, we could start thinking about it. I also need to move the core components to Helm and K8 resources

PR #1167, opened 2026-07-29
Fix jenkins agents python issues with UV

BraisVQ:
Fix python issues introduced in https://github.com/opendevstack/ods-quickstarters/releases/tag/v4.13.0 and incompativility in terraform with latest virtualenv versions (requires python 3.9)


valituguran (approved):
Looks good.

PR #1168, opened 2026-07-30
Prepare relese 4.13.1

BraisVQ:


jordivx (approved):
I would suggest putting the packages version into variables as done above (or in the tmp/requirements.txt). Since this is already approved in master and is not critical, I approve this to be released.

